PR #46, opened 2020-09-25
feat(Sample): Add DSL Java Sample

darewreck54:
Porting https://github.com/temporalio/samples-go/tree/master/dsl to java sdk

Note: Names can be better


src/main/java/io/temporal/samples/dsl/models/ActivityInvocation.java
```diff
@@ -0,0 +1,60 @@
...
+    // Class.forName(name);
+
+    ActivityStub stub =
+        Workflow.newUntypedActivityStub(
+            ActivityOptions.newBuilder()
+                .setStartToCloseTimeout(Duration.ofMinutes(1))
+                .setTaskQueue("dsl")
+                .build());
+
+    // Warning activity task takes in non-varargs call of varargs method with inexact argument type
```
darewreck54, line 38: I get this warning.  Is there another way to pass in the args?

src/main/java/io/temporal/samples/dsl/models/ActivityInvocation.java
```diff
@@ -0,0 +1,60 @@
...
+    // for last parameter;
+    // String results = stub.execute(name, String.class, args);
+    //                                                   ^
+    // cast to Object for a varargs call
+    // cast to Object[] for a non-varargs call and to suppress this warning
+    String results = stub.execute(name, String.class, args);
```
darewreck54, line 44: At this point, it executes but it doesn't the activity never returns.

Here is the stack trace
image

image
image

darewreck54: cc: @mfateev

src/main/java/io/temporal/samples/dsl/models/ActivityInvocation.java
```diff
@@ -0,0 +1,54 @@
...
+    // Class.forName(name);
+
+    ActivityStub stub =
+        Workflow.newUntypedActivityStub(
+            ActivityOptions.newBuilder()
+                .setStartToCloseTimeout(Duration.ofMinutes(1))
+                .setTaskQueue("dsl")
+                .build());
+
+    String results = stub.execute(name, String.class, new Object[] {args});
```
darewreck54, line 38: It never picks up to execute the activity.  It register the event thou
image

image

darewreck54: image

So i found out the issue. There two problems
1) It's looking for the method name of the activity (entry point) vs. the activity itself
2) in java functions are lowercase, but in this case it requires the `GetInfo` vs. `getInfo`

src/main/resources/dsl/workflow1.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+variables:
+  arg1: value1
+  arg2: value2
+
+root:
+  sequence:
+    elements:
+      - activity:
+          name: GetInfo1
```
darewreck54, line 14: Shouldn't this be the ActivityImpl's Name, vs the activity entry point method name

src/main/java/io/temporal/samples/dsl/SampleActivities.java
```diff
@@ -0,0 +1,77 @@
...
+
+import io.temporal.activity.Activity;
+import io.temporal.activity.ActivityInterface;
+import io.temporal.activity.ActivityMethod;
+
+public class SampleActivities {
+  @ActivityInterface
+  public interface SampleActivities1 {
+    @ActivityMethod
+    String getInfo1();
```
darewreck54, line 11: Is it possible to have the same method names with different Activity interfaces?  I found that if i try to registry multiple activity interfaces i can't use the same method or it conflicts

src/main/java/io/temporal/samples/dsl/models/ActivityInvocation.java
darewreck54: So the issue 1 make sense since you need to specify what method to call in the activity.  I guess i need to figure out what the code would be to generate an activity stub based on activity name...and then use the stub to call the activity method, both of which needs to be in the yaml (activity, method name)

src/main/java/io/temporal/samples/dsl/SampleActivities.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  @ActivityInterface
```
darewreck54, line 14: Ryalnd method that you can pass a prefix to add the activity name to the register method, but neither the activityInterface or the activitymethod allows for this.  Any suggestion on how this should be implemented.

I want to be able to have the same method name for different interfaces

src/main/java/io/temporal/samples/dsl/models/Parrallel.java
```diff
@@ -0,0 +1,43 @@
...
+        Workflow.newCancellationScope(
+            () -> {
+              for (Statement statement : branches) {
+                results.add(Async.function(statement::execute, bindings));
+              }
+            });
+    // As code inside the scope is non blocking the run doesn't block.
+    scope.run();
+    // If one activity fails then all the rest will fail
+    Promise.allOf(results).get();
```
mfateev, line 39: This is going to throw an exception if an activity fails. So scope.cancel should be in a catch block:
```
try {
   Promise.allOf(results).get();
} catch (RuntimeException e) {
   scope.cancel();
   throw e;
}
```

src/main/java/io/temporal/samples/dsl/models/Statement.java
```diff
@@ -0,0 +1,37 @@
+package io.temporal.samples.dsl.models;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Map;
+
+public class Statement {
+  public ActivityInvocation activity;
+  public Sequence sequence;
+  public Parrallel parrallel;
```
mfateev, line 10: I would use inheritance here assuming that Json can deserialize it.

src/main/resources/dsl/workflow1.yaml
mfateev: It has to be activity type which is in Java is defined (by default) as a method name (with the first letter capitalized) of an interface annotated with @ActivityInterface.

darewreck54:
> Porting https://github.com/temporalio/samples-go/tree/master/dsl to java sdk
> 
> Note: Names can be better

I have issue signing the CLA...it keeps taking me to a screen where i can't sign anything

src/main/java/io/temporal/samples/dsl/models/ActivityInvocation.java
darewreck54: @mfateev what would the code look like if i wanted to trigger it based on the activity name + the activities's method name.  Stub.execute("activityname":"method_name")? MethodName captalized?

mfateev:
We verified that CLA is signed by you. So it is not a blocker. I would be able to merge when it is ready.

samarabbas:
> > Porting https://github.com/temporalio/samples-go/tree/master/dsl to java sdk
> > Note: Names can be better
> 
> I have issue signing the CLA...it keeps taking me to a screen where i can't sign anything
> image

@darewreck54 you need need to include the commit email address to your github account.  Here is the response we got from cla-assistant: https://github.com/cla-assistant/cla-assistant/issues/626#issuecomment-699527544

darewreck54:
> > > Porting https://github.com/temporalio/samples-go/tree/master/dsl to java sdk
> > > Note: Names can be better
> > 
> > 
> > I have issue signing the CLA...it keeps taking me to a screen where i can't sign anything
> > image
> 
> @darewreck54 you need need to include the commit email address to your github account. Here is the response we got from cla-assistant: [cla-assistant/cla-assistant#626 (comment)](https://github.com/cla-assistant/cla-assistant/issues/626#issuecomment-699527544)

I'll need to redo this after all the comments are addressed.  Basically the original commits was using a different commit email then whats assigned on this account.  I tried to update the previous commits but it totally messed up all the commits.

mfateev:
I recommend closing this PR after addressing the comments and open a new one.

PR #47, opened 2020-09-27
[WORK IN PROGRESS] Add DSL Sample

darewreck54:
Add a dsl example.  This is to port the Go Example.   

There were issue with the previous https://github.com/temporalio/samples-java/pull/45, so created a new one.

### DIFF
--- a/src/main/java/io/temporal/samples/dsl/SampleActivities.java
+++ b/src/main/java/io/temporal/samples/dsl/SampleActivities.java
@@ -1,0 +1,77 @@
+package io.temporal.samples.dsl;
+
+import io.temporal.activity.Activity;
+import io.temporal.activity.ActivityInterface;
+import io.temporal.activity.ActivityMethod;
+
+public class SampleActivities {
+  @ActivityInterface(namePrefix = "SampleActivity1")
+  public interface SampleActivities1 {
+    @ActivityMethod(name = "SampleActivity1.getInfo")
+    String getInfo();
+  }
+
+  @ActivityInterface(namePrefix = "SampleActivity2")
+  public interface SampleActivities2 {
+    @ActivityMethod(name = "SampleActivity2.getInfo")
+    String getInfo();
+  }
+
+  @ActivityInterface(namePrefix = "SampleActivity3")
+  public interface SampleActivities3 {
+    @ActivityMethod(name = "SampleActivity3.getInfo")
+    String getInfo();
+  }
+
+  @ActivityInterface(namePrefix = "SampleActivity4")
+  public interface SampleActivities4 {
+    @ActivityMethod(name = "SampleActivity4.getInfo")
+    String getInfo();
+  }
+
+  @ActivityInterface(namePrefix = "SampleActivity5")
+  public interface SampleActivities5 {
+    @ActivityMethod(name = "SampleActivity5.getInfo")
+    String getInfo();
+  }
+
+  public static class SampleActivitiesImpl1 implements SampleActivities1 {
+    @Override
+    public String getInfo() {
+      String name = Activity.getExecutionContext().getInfo().getActivityType();
+      return "Result_" + name;
+    }
+  }
+
+  public static class SampleActivitiesImpl2 implements SampleActivities2 {
+    @Override
+    public String getInfo() {
+      String name = Activity.getExecutionContext().getInfo().getActivityType();
+      return "Result_" + name;
+    }
+  }
+
+  public static class SampleActivitiesImpl3 implements SampleActivities3 {
+    @Override
+    public String getInfo() {
+      String name = Activity.getExecutionContext().getInfo().getActivityType();
+      return "Result_" + name;
+    }
+  }
+
+  public static class SampleActivitiesImpl4 implements SampleActivities4 {
+    @Override
+    public String getInfo() {
+      String name = Activity.getExecutionContext().getInfo().getActivityType();
+      return "Result_" + name;
+    }
+  }
+
+  public static class SampleActivitiesImpl5 implements SampleActivities5 {
+    @Override
+    public String getInfo() {
+      String name = Activity.getExecutionContext().getInfo().getActivityType();
+      return "Result_" + name;
+    }
+  }
+}
--- a/src/main/java/io/temporal/samples/dsl/SampleActivities.java
+++ b/src/main/java/io/temporal/samples/dsl/SampleActivities.java
@@ -2,36 +2,30 @@ package io.temporal.samples.dsl;
 
 import io.temporal.activity.Activity;
 import io.temporal.activity.ActivityInterface;
-import io.temporal.activity.ActivityMethod;
 
 public class SampleActivities {
   @ActivityInterface(namePrefix = "SampleActivity1")
   public interface SampleActivities1 {
-    @ActivityMethod(name = "SampleActivity1.getInfo")
     String getInfo();
   }
 
   @ActivityInterface(namePrefix = "SampleActivity2")
   public interface SampleActivities2 {
-    @ActivityMethod(name = "SampleActivity2.getInfo")
     String getInfo();
   }
 
   @ActivityInterface(namePrefix = "SampleActivity3")
   public interface SampleActivities3 {
-    @ActivityMethod(name = "SampleActivity3.getInfo")
     String getInfo();
   }
 
   @ActivityInterface(namePrefix = "SampleActivity4")
   public interface SampleActivities4 {
-    @ActivityMethod(name = "SampleActivity4.getInfo")
     String getInfo();
   }
 
   @ActivityInterface(namePrefix = "SampleActivity5")
   public interface SampleActivities5 {
-    @ActivityMethod(name = "SampleActivity5.getInfo")
     String getInfo();
   }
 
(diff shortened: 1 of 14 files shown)


src/main/java/io/temporal/samples/dsl/models/Statement.java
```diff
@@ -0,0 +1,37 @@
+package io.temporal.samples.dsl.models;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Map;
+
+public class Statement {
+  public ActivityInvocation activity;
+  public Sequence sequence;
+  public Parrallel parrallel;
```
darewreck54, line 10: @mfateev You made a comment about using inheritance.  Can you give me an example of what you mean?  I wasn't following what can inherit from what?

src/main/resources/dsl/workflow1.yaml
```diff
@@ -0,0 +1,30 @@
...
+variables:
+  arg1: value1
+  arg2: value2
+
+root:
+  sequence:
+    elements:
+      - activity:
+          name: SampleActivity1
+          method: SampleActivity1.getInfo
```
darewreck54, line 15: I had to add the activity name to the method, since there is an issue with the validation registration method where it does not account for the activity name prefix when making sure no two activity/method have the same name.

src/main/java/io/temporal/samples/dsl/models/ActivityInvocation.java
```diff
@@ -0,0 +1,55 @@
...
+  public Void execute(Map<String, String> bindings) {
+    String[] args = makeInput(this.arguments, bindings);
+    ActivityStub stub =
+        Workflow.newUntypedActivityStub(
+            ActivityOptions.newBuilder()
+                .setStartToCloseTimeout(Duration.ofMinutes(5))
+                .setTaskQueue("dsl")
+                .build());
+
+    String results = stub.execute(this.method, String.class, new Object[] {args});
```
darewreck54, line 39: @mfateev  How would I use the activity's name when triggering the method.  If the activities share the same method name given that the name_prefix and method name is configured in the annotations.

darewreck54:
@mfateev I addressed the comments that I understood and added questions in the code.  If you have time, please take a look and let me know how I can proceed.  

Thanks!

src/main/java/io/temporal/samples/dsl/models/ActivityInvocation.java
mfateev, line 39: It would be nice to change activities to accept the args as the current implementation doesn't use them at all.
I'm not sure why new Object[] {args} is needed instead of just args.

src/main/java/io/temporal/samples/dsl/models/Parrallel.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @JsonCreator
+  public Parrallel(@JsonProperty("branches") Statement branches[]) {
+    this.branches = branches;
+  }
+
+  public void execute(Map<String, String> bindings) {
+    // You can use the context passed in to activity as a way to cancel the activity like standard
+    // GO way.
+    // Cancelling a parent context will cancel all the derived contexts as well.
```
mfateev, line 24: This comment doesn't make any sense to a Java developer that never learned Go.

src/main/java/io/temporal/samples/dsl/models/Statement.java
mfateev: interface Statement {}
class Sequence implements Statement {}
class ActivityInvocation implements Statement {}
class Parallel implements Statement {}

src/main/java/io/temporal/samples/dsl/models/Workflow.java
```diff
@@ -0,0 +1,8 @@
+package io.temporal.samples.dsl.models;
+
+import java.util.Map;
+
+public class Workflow {
```
mfateev, line 5: Please rename to something like DslWorkflow as Workflow class is already one of the core DSL APIs.

src/main/resources/dsl/workflow1.yaml
mfateev: The reason is that you specify the same name for all the activities. When the name is specified explicitly the prefix is ignored.

mfateev (changes requested):
Thanks a lot for the contribution!

src/main/java/io/temporal/samples/dsl/models/ActivityInvocation.java
darewreck54: If you don't do this, you get a warning in java.  The method takes a list of args which need to be expanded.  This was the workaround I found on stack-overflow to get around it.

`It would be nice to change activities to accept the args as the current implementation doesn't use them at all.`

Doesn't passing in the args like that use it in the activities?  When I look at the stack trace, it looks like it's using it as inputs in the payload.  Were you referring to something else?

image

src/main/java/io/temporal/samples/dsl/models/Statement.java
darewreck54: I'm not seeing what the interface could be.  Here is wha we have?
```
class Statement {
   public ActivityInvocation activity;
   public Sequence sequence;
   public Parallel parallel;
}

class Sequence {
  public Statement elements[];
}

class Parallel  {
  public Statement branches[];
}

class ActivityInvocation {
  public String name;
  public String method;
  public String[] arguments;
  public String result;
}
```

The only thing similar is `sequence` and `parallel`, but you would have to change the name to be consistent `elements` vs. `branches`.

src/main/resources/dsl/workflow1.yaml
darewreck54: thanks, did not know that.

src/main/java/io/temporal/samples/dsl/models/Parrallel.java
darewreck54: removed

src/main/java/io/temporal/samples/dsl/models/Workflow.java
darewreck54: Updated

src/main/java/io/temporal/samples/dsl/SampleActivities.java
```diff
@@ -0,0 +1,77 @@
+package io.temporal.samples.dsl;
+
+import io.temporal.activity.Activity;
+import io.temporal.activity.ActivityInterface;
+import io.temporal.activity.ActivityMethod;
+
+public class SampleActivities {
```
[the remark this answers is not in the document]
darewreck54: I think leaving as it shows the user how you would use the name_interface in the case a user wants to go with that approach (ex. I wanted to know and it wasn't clear to me).  But if you want me to change it, let me know.

src/main/java/io/temporal/samples/dsl/models/Statement.java
darewreck54: @mfateev ping

src/main/java/io/temporal/samples/dsl/SampleActivities.java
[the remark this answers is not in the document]
mfateev: Yes, I would change it. We can create another example that demonstrates multiple Activity interfaces.

src/main/java/io/temporal/samples/dsl/models/ActivityInvocation.java
```diff
@@ -36,7 +36,9 @@ public Void execute(Map<String, String> bindings) {
                 .setTaskQueue("dsl")
                 .build());
 
-    String results = stub.execute(this.method, String.class, new Object[] {args});
+    String methodNameDescriptor =
+        this.name + this.method.substring(0, 1).toUpperCase() + this.method.substring(1);
```
mfateev, line 40: I would use a single "activityType" string. Activities can be implemented in different languages like Go which have completely different approach to activity naming.

mfateev (changes requested):

src/main/java/io/temporal/samples/dsl/models/Parallel.java
```diff
@@ -0,0 +1,45 @@
...
+  public Statement branches[];
+
+  @JsonCreator
+  public Parallel(@JsonProperty("branches") Statement branches[]) {
+    this.branches = branches;
+  }
+
+  public void execute(Map<String, String> bindings) {
+    // In the parallel block, we want to execute all of them in parallel and wait for all of them.
+    // if one activity fails then we want to cancel all the rest of them as well.
```
flossypurse, line 23:
```suggestion
    // if one Activity Execution fails then we want to cancel all the rest of them as well.
```

tsurdilo:
This sample is fairly outdated and we are working on a new DSL example. I will close this for now and if there is any more inputs/additions you would like to add here please feel free to reopen it.

darewreck54:
@tsurdilo Sorry, i've been busy so I haven't had time to update this.  Would you like me to push the changes or do you think its not needed since it sounds like you have a new DSL example?

Also mind if i see what your new DSL example looks like?

tsurdilo:
@darewreck54 up to you I can reopen this again, np :) 
I don't have the PR ready yet, it's using DynamicWorkflow and DynamicInterface. Will ping you when its ready for review.

tsurdilo:
@darewreck54 created https://github.com/temporalio/samples-java/pull/151
will close this pr. feel free to reopen it if you like. thanks!

PR #61, opened 2020-12-04
Add sample for versioning

vitarb:
Sample that uses versioning for demo purposes.


mfateev:
Let's discuss this before proceeding as I have a lot of questions.

PR #69, opened 2021-01-29
Sample for configuring SSL enabled worker

vitarb:
Shows how to setup SSL enabled worker.
Also adds a dependency on recently added testing library.

### DIFF
--- a/src/main/java/io/temporal/samples/ssl/SslEnabledWorker.java
+++ b/src/main/java/io/temporal/samples/ssl/SslEnabledWorker.java
@@ -1,0 +1,74 @@
+/*
+ *  Copyright (c) 2020 Temporal Technologies, Inc. All Rights Reserved
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
+ *  http://aws.amazon.com/apache2.0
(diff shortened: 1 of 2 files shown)


src/main/java/io/temporal/samples/ssl/SslEnabledWorker.java
```diff
@@ -0,0 +1,74 @@
...
+    // backend.
+    // client that can be used to start and signal workflows.
+    WorkflowClient client =
+        WorkflowClient.newInstance(
+            service, WorkflowClientOptions.newBuilder().setNamespace(namespace).build());
+    // worker factory that can be used to create workers for specific task queues
+    WorkerFactory factory = WorkerFactory.newInstance(client);
+    // Worker that listens on a task queue and hosts both workflow and activity implementations.
+    Worker worker = factory.newWorker(TASK_QUEUE);
+    // TODO now register your workflow types and activity implementations.
```
Sushisource, line 69: This TODO, is it meant for the person reading the sample?

If so maybe remove the TODO and say "Now, here is where you would register... as in the below two lines"

Sushisource (approved):
LGTM small wording change

src/main/java/io/temporal/samples/ssl/SslEnabledWorker.java
vitarb: I wanted to make it more obvious that it's a required part, and didn't want to copy-paste some other sample, as it's not focused on the specific behavior. Having TODO would highlight the line in most people's IDE, making it more clear.

Sushisource: Ah, sure that makes sense. Might keep the TODO and the wording upgrade so it's a bit more clear it's a call to action.

PR #73, opened 2021-02-05
Upgrade Mockito to 3.x

GreyTeardrop:
## Overview & motivation

Update test code to Mockito 3 and adapt when needed. This PR is inspired by https://github.com/temporalio/sdk-java/issues/332 and should provide an example of how a recent Mockito can be used with Temporal SDK.

## Details

* Upgrade Mockito and PowerMock to the latest versions
* Fix deprecation warnings
* Use `withSettings().withoutAnnotations()` in `HelloActivityTest` to prevent Mockito from copying annotations from activity interface to mock class
* Re-enable `HelloChildTest.testMockedChild()` and `HelloExceptionTest.testChildWorkflowTimeout()` as they no longer fail with Mockito 3, presumably due to Mockito's switch to ByteBuddy

### DIFF
--- a/src/test/java/io/temporal/samples/hello/HelloActivityTest.java
+++ b/src/test/java/io/temporal/samples/hello/HelloActivityTest.java
@@ -90,4 +90,12 @@ public class HelloActivityTest {
     String greeting = workflow.getGreeting("World");
     assertEquals("Hello World!", greeting);
   }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testMockedActivityWithoutSettings() {
+    // Mocking activity that has method-level annotations
+    // with no withoutAnnotations() setting results in a failure
+    GreetingActivities activities = mock(GreetingActivities.class);
+    worker.registerActivitiesImplementations(activities);
+  }
 }
--- a/src/test/java/io/temporal/samples/hello/HelloActivityTest.java
+++ b/src/test/java/io/temporal/samples/hello/HelloActivityTest.java
@@ -23,6 +23,7 @@ import static io.temporal.samples.hello.HelloActivity.TASK_QUEUE;
 import static org.junit.Assert.assertEquals;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.withSettings;
 
 import io.temporal.client.WorkflowClient;
 import io.temporal.client.WorkflowOptions;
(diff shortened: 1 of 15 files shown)


GreyTeardrop:
@mfateev @vitarb I've noticed that Java SDK is getting Mockito 3 in the scope of https://github.com/temporalio/sdk-java/pull/360. Do you think it's worth upgrading samples as well?

tsurdilo:
@GreyTeardrop will need to be rebased. Also if you have time let's check if there are any examples tests that were added recently needs an update as well. Thanks!

GreyTeardrop:
Thank you, @tsurdilo! I've merged the latest `master` and updated `HelloLocalActivityTest`.

src/test/java/io/temporal/samples/hello/HelloActivityTest.java
```diff
@@ -73,7 +74,10 @@ public void testActivityImpl() {
 
   @Test
   public void testMockedActivity() {
-    GreetingActivities activities = mock(GreetingActivities.class);
+    // withoutAnnotations() is required to stop Mockito from copying
+    // method-level annotations from the GreetingActivities interface
+    GreetingActivities activities =
+        mock(GreetingActivities.class, withSettings().withoutAnnotations());
```
tsurdilo, line 80: What's the reason that you don't want Mockito to preserve the annotations? 
Why if you do preserve them (testMockedActivityWithoutSettings) you expect the exception?

Just asking because if you have an Activity method with let's say

    @ActivityMethod(name = "sayHello")
    String composeGreeting(String greeting, String name);

The activity type is different if you remove the annotation, and you probably would want to test with the 
actually registered type from the annotation name param? Same really with the @ActivityInterface annotation.

Just wondering.

GreyTeardrop: The annotation on the interface class `GreetingActivities` should not be affected. The `withoutAnnotations()` setting removes annotations that Mockito copies from the interface class to the dynamically generated mock. Without it, the `worker.registerActivitiesImplementations(activities)` call fails, as it [explicitly checks](https://github.com/temporalio/sdk-java/blob/v1.0.7/temporal-sdk/src/main/java/io/temporal/common/metadata/POJOActivityImplMetadata.java#L65-L70) that the activity implementation class should not have `@ActivityMethod` annotations.

Someone has reported that behavior as a https://github.com/temporalio/sdk-java/issues/332 issue for the Java SDK.

tsurdilo (approved):

tsurdilo:
@vitarb please review as well when you can. thanks!

vitarb (approved):

PR #82, opened 2021-03-18
Add an example for encrypting payloads.

robholland:


src/main/java/io/temporal/samples/encryptedpayloads/CryptDataConverter.java
```diff
@@ -0,0 +1,193 @@
...
+    new SecureRandom().nextBytes(nonce);
+    return nonce;
+  }
+
+  private byte[] encrypt(byte[] plainData) throws Exception {
+    SecretKey key = this.getKey();
+    byte[] nonce = getNonce(GCM_NONCE_LENGTH_BYTE);
+
+    Cipher cipher = Cipher.getInstance(CIPHER);
+    cipher.init(Cipher.ENCRYPT_MODE, key, new GCMParameterSpec(GCM_TAG_LENGTH_BIT, nonce));
```
mfateev, line 76: Is this a cheap enough operation to execute per payload?

src/main/java/io/temporal/samples/encryptedpayloads/CryptDataConverter.java
```diff
@@ -0,0 +1,193 @@
...
+    return Optional.of(encryptedPayload);
+  }
+
+  @Override
+  public <T> T fromPayload(Payload payload, Class<T> valueClass, Type valueType) {
+    ByteString encoding;
+    try {
+      encoding = payload.getMetadataOrThrow(EncodingKeys.METADATA_WRAPPED_ENCODING_KEY);
+    } catch (Throwable e) {
+      return converter.fromPayload(payload, valueClass, valueType);
```
mfateev, line 138: I wouldn't use exceptions for flow control

src/main/java/io/temporal/samples/encryptedpayloads/CryptDataConverter.java
robholland: It's required to be per payload due to the `nonce`. It must vary per encryption run.

src/main/java/io/temporal/samples/encryptedpayloads/CryptDataConverter.java
robholland: So I should use `payload.getMetadataOrDefault(...)`, or what would you suggest? I was expecting `getMetadata(key)` but that's not a thing.

robholland: I'll refactor to use `getMetadataMap` just feels weird as I only need one field.

robholland: Updated in https://github.com/temporalio/samples-java/pull/82/commits/27fbb88dbe1238ba9ffac52b85b2067a20598170

mfateev: Another option is to use `getMetadataOrDefault` which is a little cheaper than `getMetadata`.

src/main/java/io/temporal/samples/encryptedpayloads/CryptDataConverter.java
```diff
@@ -0,0 +1,189 @@
...
+      .setData(ByteString.copyFrom(plainData))
+      .removeMetadata(EncodingKeys.METADATA_WRAPPED_ENCODING_KEY)
+      .putMetadata(EncodingKeys.METADATA_ENCODING_KEY, encoding)
+      .build();
+
+    return converter.fromPayload(decryptedPayload, valueClass, valueType);
+  }
+
+  @Override
+  public Optional<Payloads> toPayloads(Object... values) throws DataConverterException {
```
mfateev, line 154: I would call `converter.toPayloads(values)` and then serialize the resulting Payload (using protobuf generated method) and then encrypt the whole thing at once.

src/main/java/io/temporal/samples/encryptedpayloads/CryptDataConverter.java
```diff
@@ -0,0 +1,189 @@
...
+    buffer.get(encryptedData);
+
+    Cipher cipher = Cipher.getInstance(CIPHER);
+    cipher.init(Cipher.DECRYPT_MODE, key, new GCMParameterSpec(GCM_TAG_LENGTH_BIT, nonce));
+
+    return cipher.doFinal(encryptedData);
+  }
+
+  @Override
+  public <T> Optional<Payload> toPayload(T value) throws DataConverterException {
```
mfateev, line 100: I don't think we want to encrypt data passed to toPayload. This method is used for searchable attributes and header fields. I don't think they should be encrypted.

src/main/java/io/temporal/samples/encryptedpayloads/CryptDataConverter.java
```diff
@@ -0,0 +1,189 @@
...
+import javax.crypto.Cipher;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+
+class EncodingKeys {
+  static final String METADATA_ENCODING_KEY = "encoding";
+  static final String METADATA_WRAPPED_ENCODING_KEY = "wrapped-encoding";
+
+  static final String METADATA_ENCODING_RAW_NAME = "binary/plain";
```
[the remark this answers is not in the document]
robholland: Updated to `binary/encrypted` per suggestion and to match go example.

src/main/java/io/temporal/samples/encryptedpayloads/CryptDataConverter.java
robholland: This didn't quite feel safe to me as Payloads has no metadata. I have however taken the suggestion of encrypting/decrypting the serialised payload rather than encrypting only the data. This is much cleaner now :)

src/main/java/io/temporal/samples/encryptedpayloads/CryptDataConverter.java
robholland: Makes sense, fixed.

src/main/java/io/temporal/samples/encryptedpayloads/CryptDataConverter.java
```diff
@@ -106,47 +110,39 @@ private SecretKey getKey() {
...
     }
 
-    Payload encryptedPayload = Payload.newBuilder(innerPayload)
-      .setData(ByteString.copyFrom(encryptedData))
-      .putMetadata(EncodingKeys.METADATA_ENCODING_KEY, EncodingKeys.METADATA_ENCODING_RAW)
-      .putMetadata(EncodingKeys.METADATA_WRAPPED_ENCODING_KEY, encoding)
-      .build();
+    Payload encryptedPayload =
+        Payload.newBuilder()
+            .putMetadata(METADATA_ENCODING_KEY, METADATA_ENCODING)
```
mfateev, line 122: I would add cipher and some sort of key id into the metadata. The idea is that the metadata should be enough to find out which key to use.

vitarb (approved):

PR #86, opened 2021-04-06
Update hello samples

tsurdilo:

## What was changed:

1) Added comments and descriptions to all of the hello examples
2) Updates examples to all include a workflow id
3) made some small updates to examples

## Why?

When I was first learning about Temporal these examples were the first thing I looked at. 
I wanted to add these changes to make it easier for people that learn from them as well
to be able to easier understand the code.

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/main/java/io/temporal/samples/hello/HelloActivity.java
+++ b/src/main/java/io/temporal/samples/hello/HelloActivity.java
@@ -20,7 +20,6 @@
 package io.temporal.samples.hello;
 
 import io.temporal.activity.ActivityInterface;
-import io.temporal.activity.ActivityMethod;
 import io.temporal.activity.ActivityOptions;
 import io.temporal.client.WorkflowClient;
 import io.temporal.client.WorkflowOptions;
@@ -48,8 +47,11 @@ public class HelloActivity {
   static final String WORKFLOW_ID = "HelloActivityWorkflow";
 
   /**
-   * Define the Workflow Interface. It must contain at least one method annotated
-   * with @WorkflowMethod
+   * Define the Workflow Interface. It must contain one method annotated with @WorkflowMethod.
+   *
+   * <p>Workflow code includes core processing logic. It that shouldn't contain any heavyweight
+   * computations, non-deterministic code, network calls, database operations, etc. All those things
+   * should be handled by Activities.
    *
    * @see io.temporal.workflow.WorkflowInterface
    * @see io.temporal.workflow.WorkflowMethod
@@ -58,16 +60,18 @@ public class HelloActivity {
   public interface GreetingWorkflow {
 
     /**
-     * Define the workflow method. This method is executed when the workflow is started. The
-     * workflow completes when the workflow method finishes execution.
+     * This method is executed when the workflow is started. The workflow completes when the
+     * workflow method finishes execution.
      */
     @WorkflowMethod
     String getGreeting(String name);
   }
 
   /**
-   * Define the Activity Interface. Workflow methods can call activities during execution.
-   * Annotating activity methods with @ActivityMethod is optional
+   * Define the Activity Interface. Activities are building blocks of any temporal workflow and
+   * contain any business logic that could perform long running computation, network calls, etc.
+   *
+   * <p>Annotating activity methods with @ActivityMethod is optional
    *
    * @see io.temporal.activity.ActivityInterface
    * @see io.temporal.activity.ActivityMethod
@@ -76,17 +80,17 @@ public class HelloActivity {
   public interface GreetingActivities {
 
     // Define your activity method which can be called during workflow execution
-    @ActivityMethod
     String composeGreeting(String greeting, String name);
   }
 
-  // Define the workflow implementation. It implements our getGreeting workflow method
+  // Define the workflow implementation which implements our getGreeting workflow method.
   public static class GreetingWorkflowImpl implements GreetingWorkflow {
 
     /**
-     * Define the GreetingActivities stub. Activity stubs implements activity interfaces and proxy
-     * calls to it to Temporal activity invocations. Since Temporal activities are reentrant, a
-     * single activity stub can be used for multiple activity invocations.
+     * Define the GreetingActivities stub. Activity stubs are proxies for activity invocations that
+     * are executed outside of the workflow thread on the activity worker, that can be on a
+     * different host. Temporal is going to dispatch the activity results back to the workflow and
+     * unblock the stub as soon as activity is completed on the activity worker.
      *
      * <p>Let's take a look at each {@link ActivityOptions} defined: The "setScheduleToCloseTimeout"
      * option sets the overall timeout that our workflow is willing to wait for activity to
@@ -104,10 +108,7 @@ public class HelloActivity {
     }
   }
 
-  /**
-   * Implementation of our workflow activity interface. It overwrites our defined composeGreeting
-   * activity method.
-   */
+  /** Simple activity implementation, that concatenates two strings. */
   static class GreetingActivitiesImpl implements GreetingActivities {
     @Override
     public String composeGreeting(String greeting, String name) {
@@ -116,15 +117,12 @@ public class HelloActivity {
   }
 
   /**
-   * With our Workflow and Activities defined, we can now start execution. The main method is our
-   * workflow starter.
+   * With our Workflow and Activities defined, we can now start execution. The main method starts
+   * the worker and then the workflow.
    */
   public static void main(String[] args) {
 
-    /*
-     * Define the workflow service. It is a gRPC stubs wrapper which talks to the docker instance of
-     * our locally running Temporal service.
-     */
+    // Define the workflow service.
     WorkflowServiceStubs service = WorkflowServiceStubs.newInstance();
 
     /*
@@ -145,8 +143,9 @@ public class HelloActivity {
     Worker worker = factory.newWorker(TASK_QUEUE);
 
     /*
-     * Register our workflow implementation with the worker. Since workflows are stateful in nature,
-     * we need to register our workflow type.
+     * Register our workflow implementation with the worker.
+     * Workflow implementations must be known to the worker at runtime in
+     * order to dispatch workflow tasks.
      */
     worker.registerWorkflowImplementationTypes(GreetingWorkflowImpl.class);
 
@@ -156,10 +155,13 @@ public class HelloActivity {
     */
     worker.registerActivitiesImplementations(new GreetingActivitiesImpl());
 
-    // Start all the workers registered for a specific task queue.
+    /*
+     * Start all the workers registered for a specific task queue.
+     * The started workers then start polling for workflows and activities.
+     */
     factory.start();
 
-    // Create our workflow client stub. It is used to start our workflow execution.
+    // Create the workflow client stub. It is used to start our workflow execution.
     GreetingWorkflow workflow =
         client.newWorkflowStub(
             GreetingWorkflow.class,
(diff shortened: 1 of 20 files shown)


src/main/java/io/temporal/samples/hello/HelloActivity.java
```diff
@@ -74,39 +104,80 @@ public String getGreeting(String name) {
...
 
-    // worker factory that can be used to create workers for specific task queues
+    /*
+     * Define the workflow factory. It is used to create workflow workers for a specific task queue.
+     */
     WorkerFactory factory = WorkerFactory.newInstance(client);
-    // Worker that listens on a task queue and hosts both workflow and activity implementations.
+
+    /*
+     * Define the workflow worker. Workflow workers listen to a defined task queue and process
```
vitarb, line 142: s/listen/listens

src/main/java/io/temporal/samples/hello/HelloActivity.java
```diff
@@ -33,34 +33,64 @@
...
+ * https://github.com/temporalio/temporal/blob/master/README.md#download-and-start-temporal-server-locally
  */
 public class HelloActivity {
 
-  static final String TASK_QUEUE = "HelloActivity";
+  // Define the task queue name
+  static final String TASK_QUEUE = "HelloActivityTaskQueue";
+
+  // Define our workflow unique id
+  static final String WORKFLOW_ID = "HelloActivityWorkflow";
```
vitarb, line 48: Should we mention about workflow id reuse policy here?

src/main/java/io/temporal/samples/hello/HelloActivity.java
```diff
@@ -33,34 +33,64 @@
...
 
-  /** Workflow interface has to have at least one method annotated with @WorkflowMethod. */
+  /**
+   * Define the Workflow Interface. It must contain at least one method annotated
```
vitarb, line 51: I believe it must have one and only one workflow method. Please double check, but I don't think we allow having multiple.

src/main/java/io/temporal/samples/hello/HelloActivity.java
```diff
@@ -33,34 +33,64 @@
...
+   * with @WorkflowMethod
+   *
+   * @see io.temporal.workflow.WorkflowInterface
+   * @see io.temporal.workflow.WorkflowMethod
+   */
   @WorkflowInterface
   public interface GreetingWorkflow {
+
+    /**
+     * Define the workflow method. This method is executed when the workflow is started. The
```
vitarb, line 61: s/Define the workflow method.// - this sentence carries no meaning, I think we should just remove it.

src/main/java/io/temporal/samples/hello/HelloActivity.java
```diff
@@ -33,34 +33,64 @@
...
+   * Annotating activity methods with @ActivityMethod is optional
+   *
+   * @see io.temporal.activity.ActivityInterface
+   * @see io.temporal.activity.ActivityMethod
+   */
   @ActivityInterface
   public interface GreetingActivities {
+
+    // Define your activity method which can be called during workflow execution
     @ActivityMethod
```
vitarb, line 79: This annotation is not required, I think we should just remove it.

src/main/java/io/temporal/samples/hello/HelloActivity.java
```diff
@@ -33,34 +33,64 @@
...
     String composeGreeting(String greeting, String name);
   }
 
-  /** GreetingWorkflow implementation that calls GreetingsActivities#composeGreeting. */
+  // Define the workflow implementation. It implements our getGreeting workflow method
```
vitarb, line 83: s/.It implements our/which implements

src/main/java/io/temporal/samples/hello/HelloActivity.java
```diff
@@ -33,34 +33,64 @@
...
+     * workflow completes when the workflow method finishes execution.
+     */
     @WorkflowMethod
     String getGreeting(String name);
   }
 
-  /** Activity interface is just a POJI. */
+  /**
+   * Define the Activity Interface. Workflow methods can call activities during execution.
+   * Annotating activity methods with @ActivityMethod is optional
```
vitarb, line 70: As mentioned below, I think we should just remove it from this sample.

src/main/java/io/temporal/samples/hello/HelloActivity.java
```diff
@@ -33,34 +33,64 @@
...
-  /** GreetingWorkflow implementation that calls GreetingsActivities#composeGreeting. */
+  // Define the workflow implementation. It implements our getGreeting workflow method
   public static class GreetingWorkflowImpl implements GreetingWorkflow {
 
     /**
-     * Activity stub implements activity interface and proxies calls to it to Temporal activity
-     * invocations. Because activities are reentrant, only a single stub can be used for multiple
-     * activity invocations.
+     * Define the GreetingActivities stub. Activity stubs implements activity interfaces and proxy
+     * calls to it to Temporal activity invocations. Since Temporal activities are reentrant, a
```
vitarb, line 88: I would replace "Activity stubs implements activity interfaces and proxy calls to it to Temporal activity invocations." with "Activity stubs are proxies for activity invocations that are executed outside of the workflow thread on the activity worker, that can be on a different host. Temporal is going to dispatch activity results back to the workflow and unblock the stub as soon as activity is completed on the activity worker."

src/main/java/io/temporal/samples/hello/HelloActivity.java
```diff
@@ -74,39 +104,80 @@ public String getGreeting(String name) {
     }
   }
 
+  /**
+   * Implementation of our workflow activity interface. It overwrites our defined composeGreeting
```
vitarb, line 108: I would rephrase:
"Simple activity implementation, that concatenates two strings."

src/main/java/io/temporal/samples/hello/HelloActivity.java
```diff
@@ -74,39 +104,80 @@ public String getGreeting(String name) {
...
   static class GreetingActivitiesImpl implements GreetingActivities {
     @Override
     public String composeGreeting(String greeting, String name) {
       return greeting + " " + name + "!";
     }
   }
 
+  /**
+   * With our Workflow and Activities defined, we can now start execution. The main method is our
+   * workflow starter.
```
vitarb, line 120: We should say that it starts the worker and then the workflow.

src/main/java/io/temporal/samples/hello/HelloActivity.java
```diff
@@ -74,39 +104,80 @@ public String getGreeting(String name) {
...
+   */
   public static void main(String[] args) {
-    // gRPC stubs wrapper that talks to the local docker instance of temporal service.
+
+    /*
+     * Define the workflow service. It is a gRPC stubs wrapper which talks to the docker instance of
```
vitarb, line 125: I would remove word docker from here, we should be able to run samples against our cloud service in the future.

src/main/java/io/temporal/samples/hello/HelloActivity.java
```diff
@@ -74,39 +104,80 @@ public String getGreeting(String name) {
...
+
+    /*
+     * Define the workflow worker. Workflow workers listen to a defined task queue and process
+     * workflows and activities.
+     */
     Worker worker = factory.newWorker(TASK_QUEUE);
-    // Workflows are stateful. So you need a type to create instances.
+
+    /*
+     * Register our workflow implementation with the worker. Since workflows are stateful in nature,
```
vitarb, line 148: 2nd sentence doesn't make much sense to me. Should we just say that workflow implementations must be known to the worker at the runtime in order to dispatch workflow tasks?

src/main/java/io/temporal/samples/hello/HelloActivity.java
```diff
@@ -33,34 +33,64 @@
...
+     * Define the workflow method. This method is executed when the workflow is started. The
+     * workflow completes when the workflow method finishes execution.
+     */
     @WorkflowMethod
     String getGreeting(String name);
   }
 
-  /** Activity interface is just a POJI. */
+  /**
+   * Define the Activity Interface. Workflow methods can call activities during execution.
```
vitarb, line 69: I would expand this a little bit.
"Activities are building blocks of any temporal workflow, and contain any business logic, that could perform long running computation, network calls, etc."

src/main/java/io/temporal/samples/hello/HelloActivity.java
vitarb, line 51: I would also say here that workflow code is core processing logic that shouldn't contain any heavyweight computation, non deterministic code, network calls, database operations, etc. All those things should live in activities.

src/main/java/io/temporal/samples/hello/HelloActivity.java
```diff
@@ -74,39 +104,80 @@ public String getGreeting(String name) {
...
-    // Activities are stateless and thread safe. So a shared instance is used.
+
+    /*
+     Register our workflow activity implementation with the worker. Since workflow activities are
+     stateless and thread-safe, we need to register a shared instance.
+    */
     worker.registerActivitiesImplementations(new GreetingActivitiesImpl());
-    // Start listening to the workflow and activity task queues.
+
+    // Start all the workers registered for a specific task queue.
```
vitarb, line 159: I would also say what happens here. Like start workers, which would start polling for workflow and activity tasks.

src/main/java/io/temporal/samples/hello/HelloActivity.java
```diff
@@ -74,39 +104,80 @@ public String getGreeting(String name) {
...
     factory.start();
 
-    // Start a workflow execution. Usually this is done from another program.
-    // Uses task queue from the GreetingWorkflow @WorkflowMethod annotation.
+    // Create our workflow client stub. It is used to start our workflow execution.
```
vitarb, line 162: nit. Here and in other places, I would remove word "our".

src/main/java/io/temporal/samples/hello/HelloActivity.java
tsurdilo: i think with "workflow workers listen" is correct (workers plural).  could change it to "A workflow worker listens" if you think that is better. let me know

src/main/java/io/temporal/samples/hello/HelloActivity.java
tsurdilo: i added a workflow id to each example so that in the web ui it does not display the unique generated one. when running multiple examples locally it helps with figuring out which workflow run is associated with what example.
can you show me maybe what this workflow reuse policy is so i can add ? you mean on like on continue run or something else ?

src/main/java/io/temporal/samples/hello/HelloActivity.java
tsurdilo: +1 yes double checked only one WorkflowMethod annotated is allowed. Will fix everywhere

tsurdilo:
@vitarb thank you for the review! updated per your comments. fixed everything except first two.

src/main/java/io/temporal/samples/hello/HelloActivity.java
Sushisource: Agree it's correct as is

vitarb (approved):

src/main/java/io/temporal/samples/hello/HelloActivity.java
vitarb: See https://docs.temporal.io/docs/concept-workflows/#workflow-id

PR #92, opened 2021-04-13
Updated HelloSearchAttributes example

tsurdilo:

## What was changed:

Updated HelloSearchAttributes:
* added comments
* added comments
* small updates to code
## Why?

This example had comments that it is not functional, but the fix has already been applied to the java sdk
Updating example so that is clear to users
## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/main/java/io/temporal/samples/hello/HelloSearchAttributes.java
+++ b/src/main/java/io/temporal/samples/hello/HelloSearchAttributes.java
@@ -52,203 +52,203 @@ import java.util.Map;
  */
 public class HelloSearchAttributes {
 
-    // Define the task queue name
-    static final String TASK_QUEUE = "HelloSearchAttributesTaskQueue";
+  // Define the task queue name
+  static final String TASK_QUEUE = "HelloSearchAttributesTaskQueue";
 
-    // Define our workflow unique id
-    static final String WORKFLOW_ID = "HelloSearchAttributesWorkflow";
+  // Define our workflow unique id
+  static final String WORKFLOW_ID = "HelloSearchAttributesWorkflow";
 
-    /**
-     * Define the Workflow Interface. It must contain one method annotated with @WorkflowMethod.
-     *
-     * <p>Workflow code includes core processing logic. It that shouldn't contain any heavyweight
-     * computations, non-deterministic code, network calls, database operations, etc. All those things
-     * should be handled by Activities.
-     *
-     * @see io.temporal.workflow.WorkflowInterface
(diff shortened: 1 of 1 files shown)


src/main/java/io/temporal/samples/hello/HelloSearchAttributes.java
```diff
@@ -165,7 +240,7 @@ private static String generateDateTimeFieldValue() {
     return ZonedDateTime.now(ZoneId.systemDefault()).format(formatter);
   }
 
-  // example for extract value from search attributes
+  // Example for extract value from search attributes
```
Sushisource, line 243: Should be changed to "extracting a value"

src/main/java/io/temporal/samples/hello/HelloSearchAttributes.java
```diff
@@ -45,33 +45,64 @@
 import java.util.UUID;
 
 /**
- * TODO(maxim): This sample is broken. https://github.com/temporalio/temporal-java-samples/issues/10
+ * Sample Temporal workflow that shows setting up and retriving workflow search attributes.
```
Sushisource, line 48: I think "demonstrates" rather than "shows" here reads a bit better

src/main/java/io/temporal/samples/hello/HelloSearchAttributes.java
```diff
@@ -85,62 +116,106 @@ public String getGreeting(String name) {
...
+  /**
+   * With our Workflow and Activities defined, we can now start execution. The main method is our
+   * workflow starter.
+   */
   public static void main(String[] args) {
-    // gRPC stubs wrapper that talks to the local docker instance of temporal service.
+    /*
+     * Define the workflow service. It is a gRPC stubs wrapper which talks to the docker instance of
+     * our locally running Temporal service.
+     */
```
Sushisource, line 138: I dunno that you need to mention "the docker instance" here. In principle it could be a full on deployment, but really the fact that it talks to the Temporal service is what matters.

Also you can probably drop "Define the" from all these comments, instead just beginning with "A gRPC stubs..."

tsurdilo:
Thanks for review @Sushisource. https://github.com/temporalio/samples-java/commit/994fab5b641b5c724488995bf1a60a5abb172e5d was merged so this is just super small pr now with just your suggested updates :)

tsurdilo:
@vitarb @Sushisource hi :) can we get approval for this please?

Sushisource (approved):

PR #95, opened 2021-04-23
Adding SideEffect example

tsurdilo:

## What was changed:

Adding SideEffect Example

## Why?

Example did not exist
## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/main/java/io/temporal/samples/hello/HelloSideEffect.java
+++ b/src/main/java/io/temporal/samples/hello/HelloSideEffect.java
@@ -39,10 +39,13 @@ import java.util.UUID;
  * Sample Temporal workflow that shows use of workflow SideEffect.
  *
  * <p>Workflow methods must be deterministic. In order to execute non-deterministic code, such as
- * random number generation as shown in this example, you should use Workflow.SideEffect
+ * random number generation as shown in this example, you should use Workflow.SideEffect.
+ * Workflow.SideEffect is typically used for very quick-running operations, where as
+ * Workflow Activities or Local Activities, which can also execute non-deterministic code,
+ * are meant for more expensive operations.
  *
  * <p>Note: you should not use SideEffect function to modify the workflow state. For that you should
- * only use the SideEffects return value!
+ * only use the SideEffect's return value!
  *
  * <p>To execute this example a locally running Temporal service instance is required. You can
  * follow instructions on how to set up your Temporal service here:
@@ -59,7 +62,7 @@ public class HelloSideEffect {
   /**
    * Define the Workflow Interface. It must contain one method annotated with @WorkflowMethod.
    *
-   * <p>Workflow code includes core processing logic. It that shouldn't contain any heavyweight
+   * <p>Workflow code includes core processing logic. It shouldn't contain any heavyweight
    * computations, non-deterministic code, network calls, database operations, etc. All those things
    * should be handled by Activities.
    *
--- a/src/main/java/io/temporal/samples/hello/HelloSideEffect.java
+++ b/src/main/java/io/temporal/samples/hello/HelloSideEffect.java
@@ -26,10 +26,11 @@ import io.temporal.client.WorkflowOptions;
 import io.temporal.serviceclient.WorkflowServiceStubs;
 import io.temporal.worker.Worker;
 import io.temporal.worker.WorkerFactory;
-import io.temporal.worker.WorkerFactoryOptions;
+import io.temporal.workflow.QueryMethod;
 import io.temporal.workflow.Workflow;
 import io.temporal.workflow.WorkflowInterface;
 import io.temporal.workflow.WorkflowMethod;
+import java.security.SecureRandom;
 import java.time.Duration;
 import java.util.Random;
 import java.util.UUID;
@@ -73,7 +74,13 @@ public class HelloSideEffect {
      * workflow method finishes execution.
      */
     @WorkflowMethod
-    String start();
+    String execute();
+
+    @QueryMethod
+    int getSafeRandomInt();
+
+    @QueryMethod
+    int getUnsafeRandomInt();
   }
 
   /**
@@ -89,7 +96,7 @@ public class HelloSideEffect {
   public interface SideEffectActivities {
 
     // Define your activity method which can be called during workflow execution
-    String doSomething(String input);
+    String greet(String greeting);
   }
 
   // Define the workflow implementation which implements our execute workflow method.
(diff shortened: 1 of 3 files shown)


src/main/java/io/temporal/samples/hello/HelloSideEffect.java
```diff
@@ -0,0 +1,266 @@
...
+     * unblock the stub as soon as activity is completed on the activity worker.
+     *
+     * <p>Let's take a look at each {@link ActivityOptions} defined: The "setScheduleToCloseTimeout"
+     * option sets the overall timeout that our workflow is willing to wait for activity to
+     * complete. For this example it is set to 2 seconds.
+     */
+    private final SideEffectActivities activities =
+        Workflow.newActivityStub(
+            SideEffectActivities.class,
+            ActivityOptions.newBuilder().setScheduleToCloseTimeout(Duration.ofSeconds(2)).build());
```
mfateev, line 111: Never use ScheduleToClose timeout in the samples. use StartToClose instead.

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
```diff
@@ -0,0 +1,266 @@
...
+      }
+
+      if (!sideEffectsRandomID.equals(savedSideEffectsRandomID)) {
+        throw new IllegalStateException("Sideffect random IDs not equal");
+      }
+
+      // This check will fail as unsafeRandomInt it was created not safely
+      // and breaks workflow determinism
+      if (unsafeRandomInt != savedUnsafeRandomInt) {
+        return "Unsafely created random numbers are not equal";
```
mfateev, line 176: The idea of "hello" samples is always to print Hello World at the end.

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
```diff
@@ -0,0 +1,266 @@
...
+    int sideEffectsRandomInt, savedSideEffectsRandomInt;
+    String sideEffectsRandomID, savedSideEffectsRandomID;
+    int unsafeRandomInt, savedUnsafeRandomInt;
+
+    @Override
+    public String start() {
+
+      // Replay-safe way to create random number using Workflow.newRandom
+      randomInt = Workflow.newRandom().nextInt();
+      savedRandomInt = Workflow.sideEffect(int.class, () -> randomInt);
```
mfateev, line 124: I don't understand why we have all these "saved..." fields.

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
```diff
@@ -0,0 +1,266 @@
...
+
+      // Replay-safe way to create random uuid
+      randomID = Workflow.randomUUID().toString();
+      savedRandomID = Workflow.sideEffect(String.class, () -> randomID);
+
+      /*
+       * Random number using side effects. Note that this value is recorded in workflow history.
+       * On replay the same value is returned so determinism is guaranteed.
+       */
+      sideEffectsRandomInt = Workflow.sideEffect(int.class, () -> new Random().nextInt());
```
mfateev, line 134: I would use https://docs.oracle.com/javase/8/docs/api/java/security/SecureRandom.html to demonstrate sideEffect.

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
```diff
@@ -0,0 +1,266 @@
...
+    /*
+     * Define the workflow factory. It is used to create workflow workers for a specific task queue.
+     * For this example we set the setWorkflowHostLocalTaskQueueScheduleToStartTimeout to zero
+     * (to disable sticky execution)
+     */
+    WorkerFactory factory =
+        WorkerFactory.newInstance(
+            client,
+            WorkerFactoryOptions.newBuilder()
+                .setWorkflowHostLocalTaskQueueScheduleToStartTimeout(Duration.ZERO)
```
mfateev, line 215: Never set this in the samples. It is purely for unit testing SDK.

In samples, you can complete a workflow and then query it which is going to force the replay.

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
```diff
@@ -0,0 +1,266 @@
...
+     * The started workers then start polling for workflows and activities.
+     */
+    factory.start();
+
+    // Create the workflow client stub. It is used to start our workflow execution.
+    SideEffectWorkflow workflow =
+        client.newWorkflowStub(
+            SideEffectWorkflow.class,
+            WorkflowOptions.newBuilder()
+                .setWorkflowId(WORKFLOW_ID)
```
mfateev, line 248: Why do you set workflow id in this sample?

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
```diff
@@ -0,0 +1,266 @@
...
+            ActivityOptions.newBuilder().setScheduleToCloseTimeout(Duration.ofSeconds(2)).build());
+
+    int randomInt, savedRandomInt;
+    String randomID, savedRandomID;
+    int sideEffectsRandomInt, savedSideEffectsRandomInt;
+    String sideEffectsRandomID, savedSideEffectsRandomID;
+    int unsafeRandomInt, savedUnsafeRandomInt;
+
+    @Override
+    public String start() {
```
mfateev, line 120: I would rename this method from start to execute or something else. The code that invokes it synchronously looks very confusing with "start" name as it waits for the workflow to complete.

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
tsurdilo: lots of samples use setScheduleToCloseTimeout. will update here and also update across examples in separate pr

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
tsurdilo: when people run samples to learn i think its easier to see which workflow run is/was associated with which example.

tsurdilo:
@mfateev thanks for the review! updated

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
tsurdilo: done

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
tsurdilo: makes sense. updated

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
tsurdilo: tried to use them for sort of comparison but queries as you said are 100 times better.

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
mfateev: I recently fixed that in Go samples: https://github.com/temporalio/samples-go/pull/89

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
```diff
@@ -0,0 +1,256 @@
...
+              () -> {
+                Random random = new SecureRandom();
+                return random.nextInt();
+              });
+
+      /*
+       * Random ID using side effects. Note that this value is recorded in workflow history.
+       * On replay the same value is returned so determinism is guaranteed.
+       */
+      sideEffectsRandomUUID = Workflow.sideEffect(String.class, UUID.randomUUID()::toString);
```
mfateev, line 148: I would not include this as it might be confusing when to use Workflow.randomUUID or the sideEffect.

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
```diff
@@ -0,0 +1,256 @@
...
+
+      /**
+       * Unsafe (not-deterministic way). We can query this after workflow execution to show that
+       * it's value changes.
+       */
+      unsafeRandomInt = new Random().nextInt();
+
+      /** Execute activity (sync) */
+      return activities.greet("World!");
```
mfateev, line 157: I would do something like:
```
String result = "Bye World!";
if ((randomUUID.version() + randomInt + sideEffectsRandomInt) % 2 == 0) {
   result = activities.greet("World");
 }
 return result;
```
Then if any of the calls is replaced with non safe random the queries are going to fail with non determinism.

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
tsurdilo: https://github.com/temporalio/samples-java/pull/96

note that i had to add retries to fileprocessingworkflow there in order to get the tests to passs. I am still trying to figure out why this is needed when the timeout changes to setStartToCloseTimeout in that case

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
```diff
@@ -0,0 +1,256 @@
...
+       * Random ID using side effects. Note that this value is recorded in workflow history.
+       * On replay the same value is returned so determinism is guaranteed.
+       */
+      sideEffectsRandomUUID = Workflow.sideEffect(String.class, UUID.randomUUID()::toString);
+
+      /**
+       * Unsafe (not-deterministic way). We can query this after workflow execution to show that
+       * it's value changes.
+       */
+      unsafeRandomInt = new Random().nextInt();
```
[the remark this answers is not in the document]
tsurdilo: removed

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
tsurdilo: thanks!

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
tsurdilo: removed

tsurdilo:
@mfateev @vitarb @Sushisource hi :) can we get approval for this please if it's ok to merge?

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
```diff
@@ -0,0 +1,247 @@
...
+import java.util.UUID;
+
+/**
+ * Sample Temporal workflow that shows use of workflow SideEffect.
+ *
+ * <p>Workflow methods must be deterministic. In order to execute non-deterministic code, such as
+ * random number generation as shown in this example, you should use Workflow.SideEffect
+ *
+ * <p>Note: you should not use SideEffect function to modify the workflow state. For that you should
+ * only use the SideEffects return value!
```
Sushisource, line 45:
```suggestion
 * only use the SideEffect's return value!
```

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
```diff
@@ -0,0 +1,247 @@
...
+import java.security.SecureRandom;
+import java.time.Duration;
+import java.util.Random;
+import java.util.UUID;
+
+/**
+ * Sample Temporal workflow that shows use of workflow SideEffect.
+ *
+ * <p>Workflow methods must be deterministic. In order to execute non-deterministic code, such as
+ * random number generation as shown in this example, you should use Workflow.SideEffect
```
Sushisource, line 42: Might be worth mentioning here why this is different from an activity or local activity, which also allow you to run nondeterministic code.

SideEffect is generally for quick-running / cheap stuff, where the activities are usually for more expensive operations

src/main/java/io/temporal/samples/hello/HelloSideEffect.java
```diff
@@ -0,0 +1,247 @@
...
+  // Define the task queue name
+  static final String TASK_QUEUE = "HelloSideEffectTaskQueue";
+
+  // Define our workflow unique id
+  static final String WORKFLOW_ID = "HelloSideEffectTaskWorkflow";
+
+  /**
+   * Define the Workflow Interface. It must contain one method annotated with @WorkflowMethod.
+   *
+   * <p>Workflow code includes core processing logic. It that shouldn't contain any heavyweight
```
Sushisource, line 62:
```suggestion
   * <p>Workflow code includes core processing logic. It shouldn't contain any heavyweight
```

Sushisource (approved):
A few small wording changes

tsurdilo:
@Sushisource thank you for the review! updated per your comments.

PR #110, opened 2021-06-10
Workflow Interceptor sample

tsurdilo:

## What was changed

Adds a simple workflow interceptor example and test

## Why?

this sample was not available

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


sw-yx (approved):

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
```diff
@@ -0,0 +1,79 @@
...
+import io.temporal.samples.interceptor.workflow.MyWorkflow;
+import io.temporal.samples.interceptor.workflow.MyWorkflowImpl;
+import io.temporal.serviceclient.WorkflowServiceStubs;
+import io.temporal.worker.Worker;
+import io.temporal.worker.WorkerFactory;
+import io.temporal.worker.WorkerFactoryOptions;
+
+public class InterceptorStarter {
+
+  public static SimpleCountWorkerInterceptor interceptor = new SimpleCountWorkerInterceptor();
```
Spikhalskiy, line 35: This is a singleton pattern which is an anti-pattern. It's better to avoid it even in the samples code.

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
```diff
@@ -0,0 +1,79 @@
...
+
+    Worker worker = factory.newWorker("test-queue");
+    worker.registerWorkflowImplementationTypes(MyWorkflowImpl.class, MyChildWorkflowImpl.class);
+    worker.registerActivitiesImplementations(new MyActivitiesImpl());
+    factory.start();
+
+    WorkflowOptions workflowOptions =
+        WorkflowOptions.newBuilder()
+            .setWorkflowId("TestWorkflow")
+            .setTaskQueue("test-queue")
```
Spikhalskiy, line 56: "test-queue" here is a text constant that is repeated in two places - a good idea to separate it in a constant.

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
```diff
@@ -0,0 +1,79 @@
...
+
+    String name = workflow.queryName();
+    String title = workflow.queryTitle();
+
+    System.out.println("Name: " + name);
+    System.out.println("Title: " + title);
+
+    try {
+      Thread.sleep(2500);
+      System.out.println(interceptor.getCountCollector().toString());
```
Spikhalskiy, line 73: It's a good idea to use a logger instead of System.out.println. It's a good idea to just make one logger config for all examples and use the looger.

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
```diff
@@ -0,0 +1,79 @@
...
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+    System.exit(0);
```
Spikhalskiy, line 77: Why do we need this? It looks like this line can be just safely omited.

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
```diff
@@ -0,0 +1,79 @@
...
+    workflow.signalNameAndTitle("John", "Customer");
+
+    String name = workflow.queryName();
+    String title = workflow.queryTitle();
+
+    System.out.println("Name: " + name);
+    System.out.println("Title: " + title);
+
+    try {
+      Thread.sleep(2500);
```
Spikhalskiy, line 72: Can we replace sleep with a wait for some event here instead? If we don't - we probably should think about adding something into SDK to make it possible.

src/main/java/io/temporal/samples/interceptor/SimpleCountWorkerInterceptor.java
```diff
@@ -0,0 +1,44 @@
...
+package io.temporal.samples.interceptor;
+
+import io.temporal.common.interceptors.ActivityInboundCallsInterceptor;
+import io.temporal.common.interceptors.WorkerInterceptor;
+import io.temporal.common.interceptors.WorkflowInboundCallsInterceptor;
+import io.temporal.samples.interceptor.collector.CountCollector;
+
+public class SimpleCountWorkerInterceptor implements WorkerInterceptor {
+
+  private CountCollector countCollector = new CountCollector();
```
Spikhalskiy, line 29: Good idea to make it a parameter and pass the instance in a constructor.

src/main/java/io/temporal/samples/interceptor/SimpleCountWorkflowInboundCallsInterceptor.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  @Override
+  public void init(WorkflowOutboundCallsInterceptor outboundCalls) {
+    super.init(new SimpleCountWorkflowOutboundCallsInterceptor(outboundCalls));
+  }
+
+  @Override
+  public WorkflowOutput execute(WorkflowInput input) {
+
```
Spikhalskiy, line 41: Did we miss counting here?

src/main/java/io/temporal/samples/interceptor/collector/CountCollector.java
```diff
@@ -0,0 +1,146 @@
...
+ */
+
+package io.temporal.samples.interceptor.collector;
+
+import io.temporal.activity.ActivityExecutionContext;
+import io.temporal.workflow.WorkflowInfo;
+import java.util.ArrayList;
+import java.util.List;
+
+public class CountCollector {
```
Spikhalskiy, line 27: CountCollector/it's internal collectons have to be thread-safe.
`InterceptorStarter.interceptor.getCountCollector().getSignalsInfoList().add(signalCollector);`
This code is/can be executed in several concurrent threads and they share the same one `CountCollector`.

src/main/java/io/temporal/samples/interceptor/workflow/MyWorkflowImpl.java
```diff
@@ -0,0 +1,60 @@
...
+  public String exec() {
+
+    // wait for a greeting
+    Workflow.await(() -> name != null && title != null);
+
+    // execute child workflow
+    MyChildWorkflow child = Workflow.newChildWorkflowStub(MyChildWorkflow.class);
+    String result = child.execChild(name, title);
+
+    Workflow.sleep(Duration.ofSeconds(2));
```
Spikhalskiy, line 40: Do we need this sleep? Is this sleep the reason why do we wait or 2500ms in InterceptorStarter? In that case, maybe we can just wait for the end of the workflow?

Spikhalskiy (changes requested):
It would be really great if code like this
`InterceptorStarter.interceptor
        .getCountCollector()
        .getActivitiesExecutionContextList()
        .add(activityExecutionContext);`
Goes away.

It's a Singleton pattern that is antipattern and should be avoided even in samples.
Most of Temporal users will be using dependency injection in their production code and dependency injection is anything but not a Singleton pattern.
My recommendation is to rewrite all classes that use CountCollector to accept it as a constructor parameter and create CountCollector explicitly in `InterceptorStarter` and pass everywhere it's needed. This will closely mimic dependency injection approach and will be a "good practice"

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
tsurdilo: Yes it is :) 
I think for this sample it's ok cause it puts focus on the interceptor and we can add comments that like you said people can use injection if use frameworks that allow it etc. But no problem, will try to update.

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
tsurdilo: Yes initially i just used logger and pass it to the interceptor classes, but that makes little sense if you have parallel execution of activities as the logs when you read them are hard to understand. So wanted to show it this way (final counts after workflow run)

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
tsurdilo: It is needed. See for example https://github.com/temporalio/samples-java/blob/master/src/main/java/io/temporal/samples/hello/HelloActivity.java#L186 and any other "starter" (main methods) in samples.

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
tsurdilo: This is not workflow code where yes you would need to use Workflow specific api for this. It's just the "starter" so its ok but I will remove it no worries.

src/main/java/io/temporal/samples/interceptor/SimpleCountWorkflowInboundCallsInterceptor.java
tsurdilo: Not really. If so it would be double count. The count for this is done in SimpleCountWorkflowOutboundCallsInterceptor

src/main/java/io/temporal/samples/interceptor/collector/CountCollector.java
tsurdilo: Yes, its just sample but will try to change to the comments.

src/main/java/io/temporal/samples/interceptor/workflow/MyWorkflowImpl.java
tsurdilo: Yeah I'll change to Workflow.await for exit signal it will be easier to understand.

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
Spikhalskiy: In my opinion, sample code should be an example of good practices. So, I would try to stick to best dev practices even if it's not the essence of the sample.

tsurdilo:
> It's a Singleton pattern that is antipattern and should be avoided even in samples.
> Most of Temporal users will be using dependency injection in their production code and dependency injection is anything but not a Singleton pattern.
> My recommendation is to rewrite all classes that use CountCollector to accept it as a constructor parameter and create CountCollector explicitly in `InterceptorStarter` and pass everywhere it's needed. This will closely mimic dependency injection approach and will be a "good practice"

Thanks for the review and comments! 
I think with Singleton its like with anything else you either hate it or hate it but see its use. For this sample I did it this way but after comments agree that there is better options so will update. I think Factory pattern probably better.

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
Spikhalskiy: 1. You don't pass loggers around in java, it's not common practice. Every class or instance just creates its own like this:
https://github.com/temporalio/sdk-java/blob/b0199150b69bb5a4eaf6c13ba4b89c2a8746a113/temporal-sdk/src/main/java/io/temporal/failure/FailureConverter.java#L50
2. I'm not saying to log from the activities, you can log it the same way as you do it here - in one place, just using a logger instead of System.out.println

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
Spikhalskiy: It's not needed and even considered bad practice. Please see: https://stackoverflow.com/a/30142476/525203
If you delete it - nothing will change.

> any other "starter" (main methods) in samples.
It's a good reason to clean up all other samples too.

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
Spikhalskiy: Yeah, I'm not saying that a Workflow class should be used here. I'm saying that it looks like this code if waiting for something. When you wait for something - you should use conditional waits (something like `while (condition is not true) sleep 100`) instead. Otherwise, there is a high chance to end up with a flaky code/test that will flake often when launched in parallels with a bunch of other tests.

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
tsurdilo: Try to run the started w/o it 
./gradlew -q execute -PmainClass=io.temporal.samples.interceptor.InterceptorStarter

or any other (see main readme) and see for yourself :)

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
tsurdilo: Yes it will be removed soon :)

src/main/java/io/temporal/samples/interceptor/SimpleCountWorkflowInboundCallsInterceptor.java
Spikhalskiy: 1. In that case what's the point in overriding a method just to make `return super.execute(input);`? We can just remove it.
2. I think it's a good idea to move the counting here for a couple of reasons:
-  you now are handling the counting of executions and signals/queries inconsistently. You count one in outbound interceptors while another in inbound interceptors. There is no reason for that, it looks just like an inconsistent implementation.
- It looks like you miss counting of the `execute` of the main root workflow. Because this would be the place to do that.

Even if you leave the current implementation - at least it should be a javadoc here about what's going on and why we count signals and queries here but don't count executions. The code itself is not self-documenting enough in this case.

tsurdilo: In WorkflowOutboundCallsInterceptorBase you can call Workflow.getInfo() which I got from your interceptor impl actually :) Can try to move to this one no problemo

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
tsurdilo: done

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
tsurdilo: done

src/main/java/io/temporal/samples/interceptor/InterceptorStarter.java
tsurdilo: done

src/main/java/io/temporal/samples/interceptor/workflow/MyWorkflowImpl.java
tsurdilo: done

Spikhalskiy (approved):

PR #129, opened 2021-07-08
Showcase JUnit5 tests and TestWorkflowExtension

GreyTeardrop:
## What was changed

This PR adds a couple of JUnit5 tests that copy existing JUnit4 tests and showcase `TestWorkflowExtension`. The PR is mostly intended as a place for discussion. Is it worth it? If it is, what's the best approach - add JUnit5 test variation for some/all existing tests? replace some existing JUnit4 tests?

Depending on opinions I could extend this PR with more tests or replace some of the existing tests with JUnit5-based ones (or something else entirely).

## Why?

This might help people using JUnit 5 to discover Temporal workflow testing tools.

## Checklist

1. Closes -
2. How was this tested: new unit tests
3. Any docs updates needed? -


tsurdilo:
@GreyTeardrop thanks for the pr! I think this looks very similar as to the junit4 TestWorkflowRule which we started using in all the tests, so changing/updating all of them imo is probably too much work for little value.
I would like to see a couple of tests that showcase TestWorkflowException as well as TestActivityExcension like what you did here, so +1 from me :)

GreyTeardrop:
Thanks, @tsurdilo! I've looked into the rest of the JUnit4 tests and it doesn't seem like porting any other of them would showcase any significantly different scenario.

As for the `TestActivityExtension`, it seems like we don't have a good scenario of it yet. The activity has to use `Activity.getExecutionContext()` otherwise it can be tested directly. Out of existing activities there's one that uses async completion (not supported by `TestActivityEnvironment` as far as I understand), several are long-running and use heart beating (not sure it's a good idea to test activities that do `Thread.sleep()`?) and some also use `Activity.getExecutionContext().getInfo().getWorkflowId()` which is always empty with `TestActivityEnvironment`.
The best option seems to be `DynamicGreetingActivityImpl` but there's an issue with `TestActivityExtension` and it doesn't support dynamic activities yet - I've included the fix into https://github.com/temporalio/sdk-java/pull/581.

tsurdilo:
@GreyTeardrop sounds good - let's then wait until your fix for https://github.com/temporalio/sdk-java/pull/581 is in and then add sample test for TestActivityExtension to this pr. WDYT?

tsurdilo:
@GreyTeardrop https://github.com/temporalio/sdk-java/pull/581 is merged from what I can tell, should we continue work on this pr?

GreyTeardrop:
> @GreyTeardrop [temporalio/sdk-java#581](https://github.com/temporalio/sdk-java/pull/581) is merged from what I can tell, should we continue work on this pr?

Thanks, @tsurdilo! 👍 I'll push some more changes soon. The CI won't probably work yet until a new Java SDK version is released, but it still makes sense from a visibility perspective.

GreyTeardrop:
Hey @tsurdilo! I've added a new test for a dynamic activity with `TestActivityExtension`

tsurdilo (approved):

PR #133, opened 2021-07-24
Disable flaky test

tsurdilo:

## What was changed

Temporarily disables HelloCancellationScopeTest
## Why?

Currently it is very flaky. Runs ok about 50% of time
Will work on fixing it but best to disable it for a bit so other prs can go through without having to re-run the build

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


vkoby:
Fixed in [Pull#137](https://github.com/temporalio/samples-java/pull/137)

PR #165, opened 2021-09-20
Sample showing use of getResult and getResultAsync

tsurdilo:

still missing update to readme and test - work in progress


src/main/java/io/temporal/samples/getresultsasync/Starter.java
```diff
@@ -0,0 +1,76 @@
...
+    untypedStub2
+        .getResultAsync(String.class)
+        .thenApply(
+            result2 -> {
+              System.out.println("Result2: " + result2);
+              return result2;
+            });
+
+    System.out.println("Waiting on result2...");
+    sleep(7); // wait 2 seconds longer to show getResultsAsync is not blocking
```
Spikhalskiy, line 64: comment is unclear. I know what you mean, but it can be misleading for the reader.

src/main/java/io/temporal/samples/getresultsasync/Starter.java
```diff
@@ -0,0 +1,76 @@
...
+    System.out.println("Done waiting on result2...");
+    System.exit(0);
```
Spikhalskiy, line 66: We really should start making a graceful shutdown in the samples instead of forcefully killing the process.

Spikhalskiy (approved):

src/main/java/io/temporal/samples/getresultsasync/Starter.java
tsurdilo: will update

PR #178, opened 2021-11-25
Remove non-deterministic call in the UpdatableTimer sample.

chronogreg:
Small tweak in the _Updatable Timer_ sample to avoid calling a non-deterministic local time zone API form the workflow.

### DIFF
--- a/src/main/java/io/temporal/samples/updatabletimer/UpdatableTimer.java
+++ b/src/main/java/io/temporal/samples/updatabletimer/UpdatableTimer.java
@@ -35,8 +35,8 @@ public final class UpdatableTimer {
 
   public void sleepUntil(long wakeUpTime) {
     Instant wakeUpInstant = Instant.ofEpochMilli(wakeUpTime);
-    LocalDateTime date = wakeUpInstant.atZone(ZoneId.systemDefault()).toLocalDateTime();
-    logger.info("sleepUntil: " + date);
+    LocalDateTime date = wakeUpInstant.atZone(ZoneId.of("UTC")).toLocalDateTime();
+    logger.info("sleepUntil: " + date + " UTC");
     this.wakeUpTime = wakeUpTime;
     while (true) {
       wakeUpTimeUpdated = false;
@@ -50,6 +50,9 @@ public final class UpdatableTimer {
   }
 
   public void updateWakeUpTime(long wakeUpTime) {
+    LocalDateTime date =
+        Instant.ofEpochMilli(wakeUpTime).atZone(ZoneId.of("UTC")).toLocalDateTime();
+    logger.info("updateWakeUpTime: " + date + " UTC");
     this.wakeUpTime = wakeUpTime;
     this.wakeUpTimeUpdated = true;
   }
--- a/src/main/java/io/temporal/samples/updatabletimer/UpdatableTimer.java
+++ b/src/main/java/io/temporal/samples/updatabletimer/UpdatableTimer.java
@@ -50,8 +50,7 @@ public final class UpdatableTimer {
   }
 
   public void updateWakeUpTime(long wakeUpTime) {
-    LocalDateTime date =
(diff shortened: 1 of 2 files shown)


src/main/java/io/temporal/samples/updatabletimer/UpdatableTimer.java
```diff
@@ -35,8 +35,8 @@
 
   public void sleepUntil(long wakeUpTime) {
     Instant wakeUpInstant = Instant.ofEpochMilli(wakeUpTime);
-    LocalDateTime date = wakeUpInstant.atZone(ZoneId.systemDefault()).toLocalDateTime();
-    logger.info("sleepUntil: " + date);
+    LocalDateTime date = wakeUpInstant.atZone(ZoneId.of("UTC")).toLocalDateTime();
```
tsurdilo, line 38: Should use ZoneOffset.UTC instead

src/main/java/io/temporal/samples/updatabletimer/UpdatableTimer.java
```diff
@@ -50,6 +50,8 @@ public void sleepUntil(long wakeUpTime) {
   }
 
   public void updateWakeUpTime(long wakeUpTime) {
+    LocalDateTime date = Instant.ofEpochMilli(wakeUpTime).atZone(ZoneId.of("UTC")).toLocalDateTime();
```
tsurdilo, line 53: same here

src/main/java/io/temporal/samples/updatabletimer/UpdatableTimer.java
tsurdilo: also can you please add comment on why this was changed? to explain users. 
raises a question, since date is just used for logging purposes, does not not make more sense to display it in users local timezon? Who knows UTC time really?

tsurdilo:
PR is also missing to update the README.md for this sample
As you have changed the output, you should also update the expected output mentioned in the readme please

tsurdilo:
Instant default timezone is utc, so this is really not needed imo. 
You can just do:

logger.info("sleepUntil: " + Instant.ofEpochMilli(wakeUpTime));
and dont need to do timezones at all

src/main/java/io/temporal/samples/updatabletimer/UpdatableTimer.java
chronogreg: Thanks, @tsurdilo. Good point, I simplified the time conversion. `Instant`, indeed, uses UTC and also includes the `Z` suffix into the `toString` value.

chronogreg: Could you please suggest more specifically what you would like to be in the comment?

You are right that in this particular instance the "side effect" does not matter, but it seems to me that in samples/tutorials, we should be particularly consistent with avoiding non-deterministic APIs, just to set a good example for the users.

In general, to make the log most readable one would want to use the local time of the _reader_, not of the _producer_. Of course, that is not as easy in this case. But using the UTC time seems like a good practice anyway.

All this is mondain, though, so not sure if this is worth mentioning in comments. Please let me know if you disagree. :)

chronogreg:
> Could you please update the README.md for this sample As proposed changes alter the output, the readme should also be updated accordingly.

Oh, sorry, @tsurdilo, I had overlooked this comment and have just rediscovered it now. 
Yes, thanks for pointing it out. ReadMe updated ([new form](https://github.com/temporalio/samples-java/blob/chronoguy/fix-updatabletimer/src/main/java/io/temporal/samples/updatabletimer/README.md)).

chronogreg:
> Instant default timezone is utc, so this is really not needed imo. You can just do:
> 
> logger.info("sleepUntil: " + Instant.ofEpochMilli(wakeUpTime)); and dont need to do timezones at all

Lol, I had overlooked this comment earlier, because I only looked at code comments by mistake. I have already shortened the time conversion based on your feedback, and it looks like I have already picked the form you suggested. ;)

tsurdilo (approved):

PR #183, opened 2021-11-29
set sw depends to 3.0.0.Final

tsurdilo:

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


tsurdilo:
setting sw depends back to 3.0.0.Final. upgrading runs into jackson issues with cloudevents depends and need to figure out the proper excludes to get this to work before upgrading.

PR #184, opened 2021-12-02
Improve the HelloPeriodic sample.

chronogreg:
The use of ContinueAsNew in the HelloPeriodic was not clear. This change aims to address that.


chronogreg:
Branching error. Will fix and reopen.

PR #185, opened 2021-12-02
Improve the HelloPeriodic sample.

chronoguy:
The use of ContinueAsNew in the HelloPeriodic was not clear. This change aims to address that.

### DIFF
--- a/src/main/java/io/temporal/samples/hello/HelloPeriodic.java
+++ b/src/main/java/io/temporal/samples/hello/HelloPeriodic.java
@@ -129,9 +129,9 @@ public class HelloPeriodic {
      * seconds.
      */
     private final GreetingActivities activities =
-            Workflow.newActivityStub(
-                    GreetingActivities.class,
-                    ActivityOptions.newBuilder().setStartToCloseTimeout(Duration.ofSeconds(10)).build());
+        Workflow.newActivityStub(
+            GreetingActivities.class,
+            ActivityOptions.newBuilder().setStartToCloseTimeout(Duration.ofSeconds(10)).build());
 
     private boolean exitRequested = false;
 
@@ -192,7 +192,7 @@ public class HelloPeriodic {
     @Override
     public void greet(String greeting) {
       System.out.println(
-              "From " + Activity.getExecutionContext().getInfo().getWorkflowId() + ": " + greeting);
+          "From " + Activity.getExecutionContext().getInfo().getWorkflowId() + ": " + greeting);
     }
   }
 
@@ -242,13 +242,13 @@ public class HelloPeriodic {
 
     // Create the workflow client stub. It is used to start our workflow execution.
     GreetingWorkflow workflow =
-            client.newWorkflowStub(
-                    GreetingWorkflow.class,
-                    // At most one instance.
-                    WorkflowOptions.newBuilder()
-                            .setWorkflowId(WORKFLOW_ID)
-                            .setTaskQueue(TASK_QUEUE)
-                            .build());
+        client.newWorkflowStub(
+            GreetingWorkflow.class,
+            // At most one instance.
+            WorkflowOptions.newBuilder()
+                .setWorkflowId(WORKFLOW_ID)
+                .setTaskQueue(TASK_QUEUE)
+                .build());
 
     // Execute our workflow.
     try {
@@ -256,8 +256,7 @@ public class HelloPeriodic {
       System.out.println("GreetingWorkflow started.");
     } catch (WorkflowExecutionAlreadyStarted e) {
       workflow = null;
-      System.out.println(
-              "GreetingWorkflow not started, because it was already running: " + e.getMessage());
+      System.out.println("GreetingWorkflow not started, because it was already running: " + e.getMessage());
     }
 
     // A temporal workflow is persistent. It will survive after this program completes.
--- a/src/main/java/io/temporal/samples/hello/HelloPeriodic.java
+++ b/src/main/java/io/temporal/samples/hello/HelloPeriodic.java
@@ -25,6 +25,7 @@ import io.temporal.activity.ActivityOptions;
 import io.temporal.client.WorkflowClient;
 import io.temporal.client.WorkflowExecutionAlreadyStarted;
 import io.temporal.client.WorkflowOptions;
+import io.temporal.client.WorkflowStub;
 import io.temporal.serviceclient.WorkflowServiceStubs;
 import io.temporal.worker.Worker;
 import io.temporal.worker.WorkerFactory;
@@ -189,6 +190,7 @@ public class HelloPeriodic {
    * With our Workflow and Activities defined, we can now start execution. The main method starts
    * the worker and then the workflow.
    */
+  @SuppressWarnings("CatchAndPrintStackTrace")  // in this simple example advanced error logging is not required
   public static void main(String[] args) throws InterruptedException {
 
     // Define the workflow service.
@@ -242,15 +244,18 @@ public class HelloPeriodic {
     // Execute our workflow.
     try {
       WorkflowClient.start(workflow::greetPeriodically, "World");
-      System.out.println("GreetingWorkflow started.");
+      System.out.println("Started a new GreetingWorkflow.");
     } catch (WorkflowExecutionAlreadyStarted e) {
-      // If the workflow is already running, we cannot start it. Instead, we will connect to the existing instance.
+      // If the workflow is already running, we cannot start it. Instead, we will connect to the
+      // existing instance.
       workflow = client.newWorkflowStub(GreetingWorkflow.class, WORKFLOW_ID);
+      System.out.println("Connected to an existing GreetingWorkflow.");
     }
 
     // The workflow is running now. In this example, we pause for a few seconds to observe its output.
     final int ObservationPeriodSecs = 20;
     System.out.println("Observing the workflow execution for " + ObservationPeriodSecs + " seconds.");
+
     try {
       Thread.sleep(ObservationPeriodSecs * 1000);
     } catch (InterruptedException intEx) {
@@ -260,6 +265,14 @@ public class HelloPeriodic {
     // Signal the workflow to exit.
     System.out.println("Requesting the workflow to exit.");
     workflow.requestExit();
+
+    // In real life we could exit now. However, in this example the workflow is running in the
+    // same process. Since workflows are persistent, if we quit before the workflow had time to
+    // react to the signal and to exit gracefully, a future invocation of this sample will find
+    // that workflow, connect to it and observe it exiting immediately.
+    // To address this, we wait for the workflow to react to the exit signal and to finish gracefully.
+    // If we run this sample again, there will be no workflow instance from a previous run left behind.
+    WorkflowStub.fromTyped(workflow).getResult(boolean.class);
 
     System.out.println("Good bye.");
     System.exit(0);
--- a/src/main/java/io/temporal/samples/hello/HelloPeriodic.java
+++ b/src/main/java/io/temporal/samples/hello/HelloPeriodic.java
@@ -97,8 +97,8 @@ public class HelloPeriodic {
     // In this example, we use an internal of 5 seconds with an intended variation of +/- 1 second
     // between executions of some useful work. In real applications a higher value may be more
     // appropriate, for example one that matches a business cycle of several hours or even days.
-    static final int SCHEDULE_PERIOD_TARGET_SECS = 5;
-    static final int SCHEDULE_PERIOD_VARIATION_SECS = 2;
+    private static final int SCHEDULE_PERIOD_TARGET_SECS = 5;
+    private static final int SCHEDULE_PERIOD_VARIATION_SECS = 2;
 
     // The max history length of a single Temporal workflow is 50,000 commands.
     // Therefore, a workflow cannot we cannot run indefinitely. Instead, we use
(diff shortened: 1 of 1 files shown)


src/main/java/io/temporal/samples/hello/HelloPeriodic.java
```diff
@@ -88,14 +96,33 @@
...
 
+    // In this example, we use an internal of 10 seconds with an intended variation of +/- 2 seconds
+    // between executions of some useful work. In real applications a higher value may be more
+    // appropriate, for example one that matches a business cycle of several hours or even days.
+    private static class ScheduleConfig {
+      static final int PeriodTargetSecs = 10;
+      static final int PeriodVariationSecs = 4;
+    }
+
+    // The max history length of a single Temporal workflow is limited.
```
tsurdilo, line 107: I really like that we add comments, but we should be clear if possible. The limitation is 50K event upon which workflow is terminated. Saying "is limited" here imo is very confusing to new users as what does it mean, 10 seconds max? Just my 2c :)

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
```diff
@@ -88,14 +96,33 @@
...
+    // Therefore, we cannot loop indefinitely. Instead, we use the ContinueAsNew
```
tsurdilo, line 108: "Therefore, we cannot loop indefinitely" - if possible please be more precise. You can loop indefinitely if in your loop you don't have any blocking operations (like workflow.sleep, activity / child workflow exec, etc). Again, super nice that we add more comments to these samples, but need to make sure that unless we are clear, then its better to not add here and let users ask us on forum/slack.

tsurdilo:
@chronogreg thanks for the PR, imho i think the changes to this sample are pretty big, to warrant it to be its own individual sample rather than a replacement for the existing one. wdyt?

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
chronogreg: Addressed. Please take a look.

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
chronogreg: Addressed. Please take a look.

chronogreg:
> @chronogreg thanks for the PR, imho i think the changes to this sample are pretty big, to warrant it to be its own individual sample rather than a replacement for the existing one. wdyt?

Good point. I made it shorter. :)

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
```diff
@@ -88,14 +94,35 @@
...
+    // affects when we need to flow to a new run: I.e. if the workflow uses more
+    // commands per iteration, then we can perform fewer iterations before we must
+    // ContinueAsNew. For a simple workflow such as this, we could perform many
+    // thousands of iterations. However, for demonstration purposes we will flow
+    // to a new run more frequently.
+    // More details: https://docs.temporal.io/docs/java/workflows/#large-event-histories
+    private static final int SingleWorkflowRunIterations = 10;
+
     // Here we introduce a random delay between periodic executions
     private final Random random = Workflow.newRandom();
```
mfateev, line 119: Add a comment that Workflow.newRandom has to be used for determinism.

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
```diff
@@ -88,14 +94,35 @@
...
+    // the ContinueAsNew feature to flow the logical execution thread to a new
+    // workflow run instance (same approach is used by Temporal's Cron-style
+    // scheduling system as well). In real life, the complexity of the workflow
+    // affects when we need to flow to a new run: I.e. if the workflow uses more
+    // commands per iteration, then we can perform fewer iterations before we must
+    // ContinueAsNew. For a simple workflow such as this, we could perform many
+    // thousands of iterations. However, for demonstration purposes we will flow
+    // to a new run more frequently.
+    // More details: https://docs.temporal.io/docs/java/workflows/#large-event-histories
+    private static final int SingleWorkflowRunIterations = 10;
```
mfateev, line 116: Java [uses](https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html) uppercase with an underscore separator for constants.

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
```diff
@@ -88,14 +94,35 @@
   // Define the workflow implementation which implements the greetPeriodically workflow method.
   public static class GreetingWorkflowImpl implements GreetingWorkflow {
 
+    // In this example, we use an internal of 5 seconds with an intended variation of +/- 1 second
+    // between executions of some useful work. In real applications a higher value may be more
+    // appropriate, for example one that matches a business cycle of several hours or even days.
+    private static class ScheduleConfig {
```
mfateev, line 100: I would remove this class as it doesn't add any value as it is just a holder of constants. Why is SingleWorkflowRunIterations are not part of any class.

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
```diff
@@ -106,31 +133,45 @@
...
-        activities.greet("Hello " + name + "! Sleeping for " + delayMillis + " milliseconds.");
-        Workflow.sleep(delayMillis);
+    public boolean greetPeriodically(String name) {
+
+      for (int i = 0; i < SingleWorkflowRunIterations; i++) {
+
+        // Compute the timing of the next iteration:
+        int delayMillis = (ScheduleConfig.PeriodTargetSecs * 1000)
+                          + random.nextInt(ScheduleConfig.PeriodVariationSecs * 1000)
+                          - (ScheduleConfig.PeriodVariationSecs * 500);
```
mfateev, line 151: I would rewrite this to use Duration class:

Duration.ofSeconds(PeriodTargetSecs).plus(Duration.ofSeconds(random.nextInt(ScheduleConfig.PeriodVariationSecs) ....

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
```diff
@@ -106,31 +133,45 @@
...
-      // Stop execution of the current workflow and start a new execution
+
+      // Create a workflow stub that will be used to continue this workflow as a new run:
+      // (see the comment for 'SingleWorkflowRunIterations' for details)
+      GreetingWorkflow continueAsNew = Workflow.newContinueAsNewStub(GreetingWorkflow.class);
+
+      // Request that the new run will be invoked by the Temporal system:
       continueAsNew.greetPeriodically(name);
+
+      return false;
```
mfateev, line 174: Why do we return boolean here? It is useless as it will only return the value of the last one to the caller which is always true.

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
chronogreg: ✔️

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
chronogreg: ✔️

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
chronogreg: ✔️

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
chronogreg: ✔️

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
chronogreg: I found this (i.e. the conversion to `Duration` at the end) shorter / easier to read.
I left it as is for now, but if you find it adds value, please let me know and I can make the change.

chronogreg:
@mfateev, thanks for the feedback. Addressed everything, except where indicated above.

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
```diff
@@ -88,14 +94,38 @@
...
 
-    // Here we introduce a random delay between periodic executions
+    // In this example, we use an internal of 5 seconds with an intended variation of +/- 1 second
+    // between executions of some useful work. In real applications a higher value may be more
+    // appropriate, for example one that matches a business cycle of several hours or even days.
+    private static final int SCHEDULE_PERIOD_TARGET_SECS = 5;
+    private static final int SCHEDULE_PERIOD_VARIATION_SECS = 2;
+
+    // The max history length of a single Temporal workflow is 50,000 commands.
+    // Therefore, a workflow cannot we cannot run indefinitely. Instead, we use
```
tsurdilo, line 104: "Therefore, a workflow cannot we cannot run indefinitely"

please fix this sentence.

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
```diff
@@ -88,14 +94,38 @@
...
+    // the ContinueAsNew feature to flow the logical execution thread to a new
```
tsurdilo, line 105: "flow the logical execution thread"
this imo is confusing, can say just "continue execution as a new run"

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
```diff
@@ -88,14 +94,38 @@
...
+    // workflow run instance (same approach is used by Temporal's Cron-style
+    // scheduling system as well). In real life, the complexity of the workflow
+    // affects when we need to flow to a new run: I.e. if the workflow uses more
+    // commands per iteration, then we can perform fewer iterations before we must
```
tsurdilo, line 109: "if the workflow uses more
commands per iteration, then we can perform fewer iterations before we must
ContinueAsNew."

imo confusing, we can just keep the comment that was there before pls?

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
```diff
@@ -88,14 +94,38 @@
...
+    // ContinueAsNew. For a simple workflow such as this, we could perform many
+    // thousands of iterations. However, for demonstration purposes we will flow
+    // to a new run more frequently.
+    // More details: https://docs.temporal.io/docs/java/workflows/#large-event-histories
+    private static final int SINGLE_WORKFLOW_ITERATIONS = 10;
+
+    // Here we introduce a random delay between periodic executions.
+    // Note that we must use 'Workflow.newRandom()' to get a new Random instance.
+    // That Random ensures that the generated numbers are uniformly randomly distributed
```
tsurdilo, line 118: "That Random ensures that the generated numbers are uniformly randomly distributed
when the workflow is executed for the first time."

imo confusing, can just say "Workflow.random() should be used inside workflow code."

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
```diff
@@ -200,11 +242,40 @@ public static void main(String[] args) throws InterruptedException {
...
+      Thread.sleep(ObservationPeriodSecs * 1000);
+    } catch (InterruptedException intEx) {
+      intEx.printStackTrace();
     }
+
+    // Signal the workflow to exit.
+    System.out.println("Requesting the workflow to exit.");
+    workflow.requestExit();
+
+    // In real life we could exit now. However, in this example the workflow is running in the
```
tsurdilo, line 270: this whole comment is confusing,

pls just say "allow workflow to finish before the worker is shut down." 

that is really the reason why workflow is still running if you exist right away, and not "to allow it to finish gracefully" 
if user started this sample back up the worker would pick up the wf task and finish it.

tsurdilo:
@chronogreg left some more comments, also the license is not signed still it seems

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
chronogreg: Sorry for the typo. 

✔️

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
chronogreg: ✔️

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
chronogreg: We changed is a few times now and I am not completely sure how exactly you prefer this. :)
I am rephrasing it to clarify the phrase in your comment. Please let me know if it works for you.

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
chronogreg: ✔️

src/main/java/io/temporal/samples/hello/HelloPeriodic.java
chronogreg: ✔️

chronogreg:
> @chronogreg left some more comments, also the license is not signed still it seems

Excellent points, @tsurdilo. Thank you. Addressed as advised, except one place, where addressed differently (indicated).
Also, I definitely had signed the license, but I just did it again. Perhaps I need to do it for each Temporal repo separately?

chronogreg:
> Also, I definitely had signed the license, but I just did it again. [...]
image

tsurdilo (approved):

PR #194, opened 2021-12-13
New scenario-based Java sample : Pizza order workflow implementation

chenjianjx:
## What was changed
New scenario-based Java sample : Pizza order workflow implementation

## Why?
Consider this as follow-up of discussion here - https://community.temporal.io/t/consider-shifting-the-focus-of-documentation-from-robustness-to-functionality/3585 .   

This can help developers understand how to implement a typical, human-interaction-involved workflow.  It also shows how the orchestration engine fits in their existing system.

## Checklist

2. How was this tested:

Follow the doc https://github.com/chenjianjx/pizza-order-workflow-demo#try-it to start up the services and invoke the business flow. 

3. Any docs updates needed?
This is about doc update only

## More words

* Feel free to change the wording
* The sample is under my own github account.  Feel free to fork it under `temporalio` account and then make changes.  That said, I hope you don't over-simplify it so that it can stay "full-fledged"

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,8 @@ All tests are available under [src/test/java](https://github.com/temporalio/samp
 
 - [**Customer Application Approval DSL**](https://github.com/temporalio/samples-java/tree/master/src/main/java/io/temporal/samples/dsl): Demonstrates execution of a customer application approval workflow defined in a DSL (like JSON or YAML)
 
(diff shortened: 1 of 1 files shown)


tsurdilo:
@chenjianjx thanks for the PR. We are not able to take it as-is, because as you mentioned the code is in your personal github repo.
Looked at your code, and looks good to me. I do think that it's a bit of a specialized sample that could definitely benefit spring boot users. Looking at it just from the java sdk sample point of view I think waiting on signals is covered also with the HelloSignal sample, but yours definitely can help people that are looking into a full, running sample using spring boot. 

We are also looking at providing spring / spring-boot support in the future, so your sample will definitely be good to have. 

One thing I would add to your sample is to show that with Temporal your workflow really becomes completely distributed. Most BPM users are used to their entire workflow running on the same host, but with temporal you could show that each activity (task) can run on a completely different service and be distributed. Just an improvement idea.

Will close this PR, but it's definitely noted. Thanks!

PR #223, opened 2022-01-31
update metrics sample

tsurdilo:

* adds custom tags to show hows that can be added
* updates starter to not shut down so users can see all the metrics on both 8080/prometheus and 8081/prometheus

### DIFF
--- a/src/main/java/io/temporal/samples/metrics/MetricsStarter.java
+++ b/src/main/java/io/temporal/samples/metrics/MetricsStarter.java
@@ -22,6 +22,7 @@ package io.temporal.samples.metrics;
 import com.sun.net.httpserver.HttpServer;
 import com.uber.m3.tally.RootScopeBuilder;
 import com.uber.m3.tally.Scope;
+import com.uber.m3.util.ImmutableMap;
 import io.micrometer.prometheus.PrometheusConfig;
 import io.micrometer.prometheus.PrometheusMeterRegistry;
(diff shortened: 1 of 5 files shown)


src/main/java/io/temporal/samples/metrics/MetricsStarter.java
```diff
@@ -38,37 +39,43 @@ public static void main(String[] args) {
...
 
-      // For the sake of the sample we shut down our process
-      // meaning we no longer record sdk metrics.
-      // In real applications, this would be a long-running process which does not
-      // shut down immediately.
-      System.exit(0);
-    } finally {
-      scrapeEndpoint.stop(1);
-    }
+    System.out.println("Starter metrics are available at http://localhost:8081/prometheus");
```
Spikhalskiy, line 75: This will be misleading. Starter JVM will finish right after this line and nothing will actually be available on 8081, while the log will state that it should be.

src/main/java/io/temporal/samples/metrics/MetricsStarter.java
```diff
@@ -38,37 +39,43 @@ public static void main(String[] args) {
...
+
+    // We don't shut down the process here so metrics can be viewed.
+    // For real applications, this would be a long-running process
```
Spikhalskiy, line 78: Can we remove everything starting with "For real applications"? No real application really looks like this starter.

Spikhalskiy (approved):

PR #247, opened 2022-03-11
Adding Crypto Data Converter Sample

tsurdilo:
Adds crypto data converter sample using https://github.com/codesqueak/jackson-json-crypto


cretz:
This seems a bit silly. Why wouldn't I encrypt the payload instead of putting encryption in the completely unrelated JSON part?

tsurdilo:
Gives ability to define what you want to encrypt or not.

Spikhalskiy (approved):

PR #275, opened 2022-04-14
Polling samples

tsurdilo:
Adds Polling samples
These samples are based on forum post: https://community.temporal.io/t/what-is-the-best-practice-for-a-polling-activity/328/2

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ All tests are available under [src/test/java](https://github.com/temporalio/samp
 
 - [**Customer Application Approval DSL**](https://github.com/temporalio/samples-java/tree/master/src/main/java/io/temporal/samples/dsl): Demonstrates execution of a customer application approval workflow defined in a DSL (like JSON or YAML)
 
-- [**Polling Services**](https://github.com/temporalio/samples-java/tree/master/src/main/java/io/temporal/samples/polling): Demonstrates best practices for polling services
+- [**Polling Services**](https://github.com/temporalio/samples-java/tree/master/src/main/java/io/temporal/samples/polling): Recommended implementation of an activity that needs to periodically poll an external resource waiting its successful completion
 
 ### API demonstrations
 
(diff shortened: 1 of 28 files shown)


tsurdilo:
@mfateev if you could take a look as well when you have time

README.md
```diff
@@ -127,6 +127,8 @@ All tests are available under [src/test/java](https://github.com/temporalio/samp
 
 - [**Customer Application Approval DSL**](https://github.com/temporalio/samples-java/tree/master/src/main/java/io/temporal/samples/dsl): Demonstrates execution of a customer application approval workflow defined in a DSL (like JSON or YAML)
 
+- [**Polling Services**](https://github.com/temporalio/samples-java/tree/master/src/main/java/io/temporal/samples/polling): Demonstrates best practices for polling services
```
Spikhalskiy, line 130: I would probably change the wording. It's hardly can be called "best practice", because it's elegant, but a hack.
Maybe we should call it "a technique"?

src/main/java/io/temporal/samples/polling/PollingChildWorkflow.java
```diff
@@ -0,0 +1,29 @@
...
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.samples.polling;
+
+import io.temporal.workflow.WorkflowInterface;
+import io.temporal.workflow.WorkflowMethod;
+
+@WorkflowInterface
+public interface PollingChildWorkflow {
```
Spikhalskiy, line 26: Do we really really need child workflows for this samples? Can we make it simpler?

src/main/java/io/temporal/samples/polling/README.md
```diff
@@ -0,0 +1,9 @@
+# Polling
+
+These samples show three different best practices for polling.
+
+1. [Frequent Polling](frequent/README.md)
+2. [Infrequent Polling](infrequent/README.md)
+3. [Periodic Polling](periodic/README.md)
```
Spikhalskiy, line 7: I think this structure is very confusing. I think it should by "Frequently Polling Activity" "Infrequently Polling Activity" and "Periodic execution on a sequence of activities"

PR #360, opened 2022-09-29
added replay-test

antmendoza:
## What was changed

Added test for `WorkflowReplayer` functionality 

## Why?

There is no a example of `WorkflowReplayer` in this repository


tsurdilo:
I think instead of hard-coding the json you could run the workflow in test and then get its history to feed into replayer, for example:

         testWorkflowRule.getHistory(execution)).toJson(true)

antmendoza:
ok, I am going to change it

PR #393, opened 2022-12-05
use versioning inside loop

antmendoza:
## What was changed

This example illustrates how to version code inside a loop.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/main/java/io/temporal/samples/hello/HelloVersioningCodeInALoop.java
+++ b/src/main/java/io/temporal/samples/hello/HelloVersioningCodeInALoop.java
@@ -174,10 +174,7 @@ public class HelloVersioningCodeInALoop {
     @WorkflowMethod
     void execute();
 
-
-    /**
-     * Signal the workflow that is waiting inside the loop to execute to continue.
-     */
+    /** Signal the workflow that is waiting inside the loop to continue. */
     @SignalMethod
     void executeNextIteration();
 
--- a/src/main/java/io/temporal/samples/hello/HelloVersioningCodeInALoop.java
+++ b/src/main/java/io/temporal/samples/hello/HelloVersioningCodeInALoop.java
@@ -1,0 +1,301 @@
+/*
+ *  Copyright (c) 2020 Temporal Technologies, Inc. All Rights Reserved
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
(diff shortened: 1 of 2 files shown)


cretz:
@antmendoza - Is there anything novel about this example compared to regular versioning except that we just added a "for" loop around the versioning? Just curious if there is something special I am missing, or if this is like "versioning inside `if` statement", "versioning inside `switch` stateement", etc. Our goal of course is to make versioning work regardless of construct.

antmendoza:
@cretz you are probably right. I added this because I thought it would be useful. 

How would you introduce a change here? 
https://github.com/temporalio/samples-java/pull/393/files#diff-484b9d9fac50b30420ebd828b74addcc67bf5ea3cf2b029a044bcc1f6ad34dfcR208-R212

cretz:
> How would you introduce a change here?

Depends on what kind of change. But in general, I'd put _any_ new code behind a new version. Technically if your use case is "half way through the loop do something new even on old workflows" then your approach of change-id-each-iteration technically works. But you'll add a lot of unnecessary version markers for new workflows. I'd probably do _close_ to what you have in this PR but stop calling get version after first seen:

```java
// Example of a loop that does a new thing on code change
int seenNewVersionOnIter = -1;
for (int i = 0; i < 200; i++) {
  // Use the current iteration until we see a version change
  String changeId = "changeId-" + (seenNewVersionOnIter < 0 ? i : seenNewVersionOnIter);
  int version = Workflow.getVersion(changeId, Workflow.DEFAULT_VERSION, 1);
  if (version == Workflow.DEFAULT_VERSION) {
    executeActivityPerformX();
  } else {
    seenNewVersionOnIter = i;
    executeActivityPerformY();
  }
  
  // ...
}
```

But this is a bit advanced and a rare use case. And I'll defer to @Spikhalskiy at this point to see if there's a better pattern and whether we want to encourage the pattern I pasted.

tsurdilo:
Hi, is this ok to merge?
Also do you think we should have a test for this a well?

tsurdilo:
If we add this sample idk maybe would also be nice to add WorkflowReplayer test as well somehow?

Spikhalskiy:
Honestly, I don't really understand what's going on here. And I think the reader/user will not get it too.
1. This needs a better description.
2. Maybe I'm not getting something, but why do we change the `changeId` on every execution of the loop? I think using one value should be just fine here.

> Using the same changeId will not work as it is cached, and is going to return -1 always.
It doesn't sound right. Are you sure?

tsurdilo:
@antmendoza hi, based on comments here do you think we can close the pr?

antmendoza:
@tsurdilo sure ,  I am closing it, no problem

PR #413, opened 2023-01-18
Retry activity on signal interceptor

mfateev:
Demonstrates how to create and register an interceptor that retries an activity on a signal.

### DIFF
--- a/src/main/java/io/temporal/samples/retryonsignalinterceptor/MyActivityImpl.java
+++ b/src/main/java/io/temporal/samples/retryonsignalinterceptor/MyActivityImpl.java
@@ -26,6 +26,7 @@ public class MyActivityImpl implements MyActivity {
 
   private final AtomicInteger count = new AtomicInteger();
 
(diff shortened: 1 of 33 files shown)


src/main/java/io/temporal/samples/retryonsignalinterceptor/MyActivityImpl.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package io.temporal.samples.retryonsignalinterceptor;
+
+import io.temporal.failure.ApplicationFailure;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class MyActivityImpl implements MyActivity {
+
+  private final AtomicInteger count = new AtomicInteger();
```
bergundy, line 27: I would look at the activity info attempt instead.
The way it is currently implemented will of course not work in a "real" activity since it may be executed on different workers and I'd rather not encourage this practice.

bergundy (approved):

PR #455, opened 2023-05-26
Sample - simple payload coded use to encode/decode failure messages

tsurdilo:
Sample that shows how to set up simple payload codec to encode/decode failure messages

### DIFF
--- a/src/main/java/io/temporal/samples/encodefailures/Starter.java
+++ b/src/main/java/io/temporal/samples/encodefailures/Starter.java
@@ -44,11 +44,11 @@ public class Starter {
     // sets encodeFailureAttributes to true
     CodecDataConverter codecDataConverter =
         new CodecDataConverter(
-            // for sampel we just use default data converter
+            // For sample we just use default data converter
             DefaultDataConverter.newDefaultInstance(),
-            // simple prefix codec to encode/decode
+            // Simple prefix codec to encode/decode
             Collections.singletonList(new SimplePrefixPayloadCodec()),
-            true); // setting encodeFailureAttributes to true
+            true); // Setting encodeFailureAttributes to true
 
     // WorkflowClient uses our CodecDataConverter
     WorkflowClient client =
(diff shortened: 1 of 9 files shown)


src/main/java/io/temporal/samples/encodefailures/Starter.java
```diff
@@ -0,0 +1,110 @@
...
+  private static final String WORKFLOW_ID = "CustomerValidationWorkflow";
+
+  public static void main(String[] args) {
+    WorkflowServiceStubs service = WorkflowServiceStubs.newLocalServiceStubs();
+
+    // CodecDataConverter defines our data converter and codec
+    // sets encodeFailureAttributes to true
+    CodecDataConverter codecDataConverter =
+        new CodecDataConverter(
+            // for sampel we just use default data converter
```
Quinn-With-Two-Ns, line 47: `sample`

src/main/java/io/temporal/samples/encodefailures/Starter.java
```diff
@@ -0,0 +1,110 @@
...
+        client.newWorkflowStub(
+            CustomerAgeCheck.class,
+            WorkflowOptions.newBuilder()
+                .setWorkflowId(WORKFLOW_ID)
+                .setTaskQueue(TASK_QUEUE)
+                .build());
+
+    try {
+      // Start workflow execution to validate under-age customer
+      workflow.validateCustomer(new MyCustomer("John", 17));
```
Quinn-With-Two-Ns, line 73: If the workflow does not fail should we log a failure since our assertion failed?

Quinn-With-Two-Ns (approved):

src/main/java/io/temporal/samples/encodefailures/Starter.java
tsurdilo: fixed, thanks

src/main/java/io/temporal/samples/encodefailures/Starter.java
tsurdilo: added a printout

PR #464, opened 2023-06-06
SpringBoot sample (and project refactor)

tsurdilo:
Refactors samples to multi-project (core and springboot).
Adds SpringBoot "hello" sample (more to be added soon if this gets in).


springboot/src/main/resources/application.yaml
```diff
@@ -0,0 +1,25 @@
+server:
+  port: 3030
+spring:
+  application:
+    name: temporal-samples
+  temporal:
+    connection:
+      target: 127.0.0.1:7233
+      target.namespace: default
+#    workers:
```
Quinn-With-Two-Ns, line 10: Any reason to leave these in?

tsurdilo: would it be ok to leave it and put a comment there "here are some other options you could set"? 
we might not have an example for some of these options and users often ask us how to set them

Quinn-With-Two-Ns: Yeah totally fine with having some example options not set by default. Just a comment explaining they are intentionally commented out it fine. Maybe we could also include TLS configuration through spring boot options?

tsurdilo: Great idea

Quinn-With-Two-Ns (approved):
LGTM though I admit I didn't try to run the individual samples

core/src/main/java/io/temporal/samples/asyncchild/README.md
cretz, line 1: Are all of these READMEs still correct? Meaning the line:

```
./gradlew -q execute -PmainClass=io.temporal.samples.asyncchild.Starter
```

still works when executed from the root?

README.md
```diff
@@ -1,192 +1,125 @@
...
-  - [Temporal Web UI](#temporal-web-ui)
-  - [Running samples](#running-samples)
-  - [Samples directory](#samples-directory)
-    - [Hello samples](#hello-samples)
-    - [Scenario-based samples](#scenario-based-samples)
-    - [API demonstrations](#api-demonstrations)
-    - [SDK Metrics](#sdk-metrics)
-    - [Tracing Support](#tracing-support)
-  - [IDE Integration](#ide-integration)
-    - [IntelliJ](#intellij)
```
cretz, line 23: Seems we lost the link to the server repo, the SDK repo, the docsm, and the TOC. Is this by intention?

README.md
```diff
@@ -1,192 +1,125 @@
...
-
-## Samples directory
-
-The following section lists all available samples.
-Click on the sample link to view the README, which contains instructions on how to run them.
-
-Each sample has an associated unit test which demonstrates the use of the Temporal Java SDK testing API.
-All tests are available under [src/test/java](./src/test/java/io/temporal/samples)
+### Running "Core" samples
+See the README.md file in each main sample directory for cut/paste Gradle command to run specific example.
```
cretz, line 35: There needs to be somewhere where we tell them to run a local server. We don't have to say how (e.g. the docker compose stuff you deleted), but we should link docs or something.

core/.gitignore
cretz, line 1: Would prefer we just keep a top-level `.gitignore` and alter the absolute path ones if needed

tsurdilo: ok will update. ty

tsurdilo: done

README.md
tsurdilo: done

README.md
tsurdilo: done

cretz (approved):
LGTM

PR #469, opened 2023-06-08
Add sample showing sync update

Quinn-With-Two-Ns:
Add a sample showing how to use the new sync update feature

### DIFF
--- a/core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
+++ b/core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
@@ -1,0 +1,281 @@
+/*
+ *  Copyright (c) 2020 Temporal Technologies, Inc. All Rights Reserved
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.samples.hello;
+
+import com.google.common.base.Throwables;
+import io.temporal.activity.ActivityOptions;
+import io.temporal.client.WorkflowClient;
+import io.temporal.client.WorkflowOptions;
+import io.temporal.client.WorkflowStub;
+import io.temporal.client.WorkflowUpdateException;
+import io.temporal.failure.ApplicationFailure;
+import io.temporal.serviceclient.WorkflowServiceStubs;
+import io.temporal.worker.Worker;
+import io.temporal.worker.WorkerFactory;
+import io.temporal.workflow.SignalMethod;
+import io.temporal.workflow.UpdateMethod;
+import io.temporal.workflow.UpdateValidatorMethod;
+import io.temporal.workflow.Workflow;
+import io.temporal.workflow.WorkflowInterface;
+import io.temporal.workflow.WorkflowMethod;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Sample Temporal workflow that demonstrates how to use workflow update methods to update a
+ * workflow execution from external sources.
+ *
+ * <p>Note: Make sure to set {@code frontend.enableUpdateWorkflowExecution=true} in you Temporal
+ * config to enabled update.
+ */
+public class HelloUpdate {
+
+  // Define the task queue name
+  static final String TASK_QUEUE = "HelloUpdateTaskQueue";
+
+  // Define the workflow unique id
+  static final String WORKFLOW_ID = "HelloUpdateWorkflow";
+
+  /**
+   * The Workflow Definition's Interface must contain one method annotated with @WorkflowMethod.
+   *
+   * <p>Workflow Definitions should not contain any heavyweight computations, non-deterministic
+   * code, network calls, database operations, etc. Those things should be handled by the
+   * Activities.
+   *
+   * @see WorkflowInterface
+   * @see WorkflowMethod
+   */
+  @WorkflowInterface
+  public interface GreetingWorkflow {
+    /**
+     * This is the method that is executed when the Workflow Execution is started. The Workflow
+     * Execution completes when this method finishes execution.
+     */
+    @WorkflowMethod
+    List<String> getGreetings();
+
+    /*
+     * Define the workflow waitForName update method. This method is executed when the workflow
+     * receives an update request.
+     */
+    @UpdateMethod()
+    Integer waitForName(String name);
+
+    /*
+     * Define an optional workflow update validator. The validator must take the same parameters as the update handle.
+     * The validator is run before the update handle.
+     * If the validator fails by throwing any exception the update request will be rejected and the handle will not run.
+     */
+    @UpdateValidatorMethod(updateName = "waitForName")
+    void waitForNameValidator(String name);
+
+    // Define the workflow exit signal method. This method is executed when the workflow receives a
+    // signal.
+    @SignalMethod
+    void exit();
+  }
+
+  // Define the workflow implementation which implements the getGreetings workflow method.
+  public static class GreetingWorkflowImpl implements GreetingWorkflow {
+
+    // messageQueue holds up to 10 messages (received from updates)
+    List<String> messageQueue = new ArrayList<>(10);
+    List<String> receivedMessages = new ArrayList<>(10);
+    boolean exit = false;
+
+    private final HelloActivity.GreetingActivities activities =
+        Workflow.newActivityStub(
+            HelloActivity.GreetingActivities.class,
+            ActivityOptions.newBuilder().setStartToCloseTimeout(Duration.ofSeconds(2)).build());
+
+    @Override
+    public List<String> getGreetings() {
+
+      while (true) {
+        // Block current thread until the unblocking condition is evaluated to true
+        Workflow.await(() -> !messageQueue.isEmpty() || exit);
+        if (messageQueue.isEmpty() && exit) {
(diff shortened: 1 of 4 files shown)


tsurdilo (approved):

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
```diff
@@ -0,0 +1,281 @@
...
+     * Execution completes when this method finishes execution.
+     */
+    @WorkflowMethod
+    List<String> getGreetings();
+
+    /*
+     * Define the workflow waitForName update method. This method is executed when the workflow
+     * receives an update request.
+     */
+    @UpdateMethod()
```
cretz, line 80:
```suggestion
    @UpdateMethod
```

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
```diff
@@ -0,0 +1,281 @@
...
+        // Block current thread until the unblocking condition is evaluated to true
+        Workflow.await(() -> !messageQueue.isEmpty() || exit);
+        if (messageQueue.isEmpty() && exit) {
+          // no messages in queue and exit signal was sent, return the received messages
+          return receivedMessages;
+        }
+        String message = messageQueue.remove(0);
+        receivedMessages.add(message);
+      }
+    }
```
cretz, line 123: This is clearer to me:

```java
while (!exit) {
  Workflow.await(() => !messageQueue.isEmpty() || exit);
  receivedMessages.addAll(messageQueue);
  messageQueue.clear();
}
return receivedMessages;
```

Subjective of course

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
```diff
@@ -0,0 +1,281 @@
...
+     */
+    @WorkflowMethod
+    List<String> getGreetings();
+
+    /*
+     * Define the workflow waitForName update method. This method is executed when the workflow
+     * receives an update request.
+     */
+    @UpdateMethod()
+    Integer waitForName(String name);
```
cretz, line 81:
```suggestion
    int waitForName(String name);
```

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
```diff
@@ -0,0 +1,281 @@
...
+          // no messages in queue and exit signal was sent, return the received messages
+          return receivedMessages;
+        }
+        String message = messageQueue.remove(0);
+        receivedMessages.add(message);
+      }
+    }
+
+    @Override
+    public Integer waitForName(String name) {
```
cretz, line 126: What does `waitForName` mean? Not sure "wait" is the obvious term for what is happening. Maybe just `addGreeting`?

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
```diff
@@ -0,0 +1,281 @@
...
+      if (name.isEmpty()) {
+        // Updates can fail by throwing a TemporalFailure
+        throw ApplicationFailure.newFailure("Cannot greet someone with an empty name", "Failure");
+      }
+      // Updates can mutate workflow state like variables or call activities
+      messageQueue.add(activities.composeGreeting("Hello", name));
```
cretz, line 132: May need to make it clear here or somewhere that if `exit` is received while the update is still running, the update will fail (or can rework your code to disallow exits while updates are running)

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
```diff
@@ -0,0 +1,281 @@
...
+    @Override
+    public void waitForNameValidator(String name) {
+      /*
+       * Update validators have the same restrictions as Queries. So workflow state cannot be
+       * mutated inside a query.
+       */
+      if (receivedMessages.size() >= 10) {
+        /*
+         * Throwing an exception inside an update validator will cause the update to be rejected.
+         * Note: rejected update will not appear in the workflow history
```
cretz, line 146: Maybe a note why `IllegalStateException` can be used here but `IllegalArgumentException` can't be used in the updater

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
```diff
@@ -0,0 +1,281 @@
...
+         */
+        throw new IllegalStateException("Only 10 greetings may be added to the queue");
```
cretz, line 148:
```suggestion
        throw new IllegalStateException("Only 10 greetings may be added");
```

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
```diff
@@ -0,0 +1,281 @@
...
+    WorkflowClient.start(workflow::getGreetings);
+
+    // After start for getGreeting returns, the workflow is guaranteed to be started.
+    // So we can send an update to it using the workflow stub.
+    // This workflow keeps receiving updates until exit is called
+
+    // When the workflow is started the getGreetings will block for the previously defined
+    // conditions
+    // Send the first workflow update
+    workflow.waitForName("World");
```
cretz, line 219: You demonstrate returning from an update but don't ever use the result

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
```diff
@@ -0,0 +1,281 @@
...
+     * This reconnects our workflowById workflow stub to the existing workflow and blocks until
+     * a result is available. Note that this behavior assumes that WorkflowOptions are not configured
+     * with WorkflowIdReusePolicy.AllowDuplicate. If they were, this call would fail with the
+     * WorkflowExecutionAlreadyStartedException exception.
+     */
+    List<String> greetings = workflowById.getGreetings();
+
+    // Print our two greetings which were sent by signals
+    System.out.println(greetings);
+    System.exit(0);
```
cretz, line 279:
```suggestion
```

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
```diff
@@ -0,0 +1,281 @@
...
+    void exit();
+  }
+
+  // Define the workflow implementation which implements the getGreetings workflow method.
+  public static class GreetingWorkflowImpl implements GreetingWorkflow {
+
+    // messageQueue holds up to 10 messages (received from updates)
+    List<String> messageQueue = new ArrayList<>(10);
+    List<String> receivedMessages = new ArrayList<>(10);
+    boolean exit = false;
```
cretz, line 103:
```suggestion
    private final List<String> messageQueue = new ArrayList<>(10);
    private final List<String> receivedMessages = new ArrayList<>(10);
    private boolean exit;
```

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
Quinn-With-Two-Ns: The  result is used, just not by every call.

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
Quinn-With-Two-Ns: I had this since all the other samples I saw had it, not sure why TBH. Don't think we run any `daemon` threads

mmcshane (approved):
couple of notes, not blocking

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
```diff
@@ -0,0 +1,281 @@
...
+          return receivedMessages;
+        }
+        String message = messageQueue.remove(0);
+        receivedMessages.add(message);
+      }
+    }
+
+    @Override
+    public Integer waitForName(String name) {
+      if (name.isEmpty()) {
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yeah just to show how to fail an update, I'll add a note

cretz (approved):
No blocking things left from my POV

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
Quinn-With-Two-Ns: Yes that's a  good point. I'll note that an in flight update may not stop workflow completion.

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
```diff
@@ -0,0 +1,287 @@
...
+     * Define the workflow addGreeting update method. This method is executed when the workflow
+     * receives an update request.
+     */
+    @UpdateMethod
+    int addGreeting(String name);
+
+    /*
+     * Define an optional workflow update validator. The validator must take the same parameters as the update handle.
+     * The validator is run before the update handle.
+     * If the validator fails by throwing any exception the update request will be rejected and the handle will not run.
```
mfateev, line 86: I would mention that the update is not recorded in the history if thrown from the validator and is recorded if thrown from the update method itself.

mfateev: Also, we need to document that validator must never change the workflow state (the same as query method).

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
```diff
@@ -0,0 +1,287 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Sample Temporal workflow that demonstrates how to use workflow update methods to update a
+ * workflow execution from external sources.
+ *
+ * <p>Note: Make sure to set {@code frontend.enableUpdateWorkflowExecution=true} in your Temporal
+ * config to enabled update.
+ */
```
mfateev, line 48: I don't like this example. It doesn't demonstrate why an update should is better than a signal. 
I would make it simpler to remove the queue and update some workflow state, rejecting it if it was updated from a different client. Some version value can drive the rejection.

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
Quinn-With-Two-Ns: Yes, both of those are documented in this sample

core/src/main/java/io/temporal/samples/hello/HelloUpdate.java
Quinn-With-Two-Ns: The update handle does modify workflow state. I think it also does show the benefit of update over a query or a signal because it modifies the workflow state like a signal and it returns a result like a query. That being said it is probably  worth calling out these benefits explicitly here as  well.

PR #471, opened 2023-06-15
Add sample showing schedules API

Quinn-With-Two-Ns:
## What was changed
Add sample showing schedules API based on the Go sample

## Why?
Show users how to use the new schedules API


antmendoza:
closes #471

Quinn-With-Two-Ns:
Oh yes I missed that readme,  I'll add it there

PR #490, opened 2023-06-27
SpringBoot - Sync Update Example

tsurdilo:
Sync Update SpringBoot Sample

This sample shows how to use update feature. Uses jpa and h2 in-memory db for the inventory. 
User can enter a fishing item to purchase from inventory and define the amount.
The update validator checks if inventory has item and the amount (local activity)
The update then updates the db (local activity).

If amount you wan to purchase is > than whats in inventory the validator fails and shows message.
On success purchase the inventory table is updated dynamically..

Screen Shot 2023-06-26 at 11 43 59 PM

(this pr also includes small changes to other sample html pages, like using fragments for header and footer)

### DIFF
--- a/springboot/src/main/java/io/temporal/samples/springboot/SamplesController.java
+++ b/springboot/src/main/java/io/temporal/samples/springboot/SamplesController.java
@@ -102,10 +102,13 @@ public class SamplesController {
 
     // send update
     try {
-      workflow.makePurchase(purchase);
+      boolean isValidPurchase = workflow.makePurchase(purchase);
       // for sample send exit to workflow exec and wait till it completes
       workflow.exit();
       WorkflowStub.fromTyped(workflow).getResult(Void.class);
+      if (!isValidPurchase) {
+        return new ResponseEntity("\"Invalid purchase\"", HttpStatus.NOT_FOUND);
+      }
       return new ResponseEntity("\"" + "Purchase successful" + "\"", HttpStatus.OK);
     } catch (WorkflowUpdateException | StatusRuntimeException e) {
       // for sample send exit to workflow exec and wait till it completes
--- a/springboot/src/main/java/io/temporal/samples/springboot/SamplesController.java
+++ b/springboot/src/main/java/io/temporal/samples/springboot/SamplesController.java
@@ -19,10 +19,16 @@
 
 package io.temporal.samples.springboot;
 
+import io.grpc.StatusRuntimeException;
 import io.temporal.client.WorkflowClient;
 import io.temporal.client.WorkflowOptions;
+import io.temporal.client.WorkflowStub;
+import io.temporal.client.WorkflowUpdateException;
 import io.temporal.samples.springboot.hello.HelloWorkflow;
 import io.temporal.samples.springboot.hello.model.Person;
+import io.temporal.samples.springboot.update.PurchaseWorkflow;
+import io.temporal.samples.springboot.update.model.ProductRepository;
+import io.temporal.samples.springboot.update.model.Purchase;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
@@ -35,6 +41,8 @@ import org.springframework.web.bind.annotation.*;
 public class SamplesController {
 
   @Autowired WorkflowClient client;
+
+  @Autowired ProductRepository productRepository;
 
   @GetMapping("/hello")
   public String hello(Model model) {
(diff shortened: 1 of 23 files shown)


springboot/src/main/java/io/temporal/samples/springboot/update/PurchaseWorkflowImpl.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  @Override
+  public boolean makePurchase(Purchase purchase) {
+    return activities.makePurchase(purchase);
+  }
+
+  @Override
+  public void makePurchaseValidator(Purchase purchase) {
+    if (!activities.isProductInStockForPurchase(purchase)) {
```
Quinn-With-Two-Ns, line 51: Local activities cannot be run in validators

Quinn-With-Two-Ns: they have the same restriction as queries

tsurdilo: it works as in it gets called ok. didnt see failure

Quinn-With-Two-Ns: it is not supported and It will not replay correctly

tsurdilo: if that is enforced tho idk why? would be very restrictive if could not check db for example to validate update

tsurdilo: here is event history that shows that local activity called in validator does get executed if that helps.
this is the path when you enter in amount thats greater than that in inventory:

Screen Shot 2023-06-27 at 1 36 10 AM

tsurdilo: ill add WorkflowReplayer test for this tomorrow and see 

but if it does not work i think the validators are pretty weak honestly. for use cases like this one not being able to check db or invoke some api to validate is imho useless

Quinn-With-Two-Ns: validators cannot do anything that modifies workflow state like executing an activity or modifying a workflow variable.

They are only able to read workflow state and update inputs. There is desire to support running local activities in update validators, but the server does not support this yet.

tsurdilo: ok will move that check to update method then. so if it does nit update wf state its ok like in this example?

Quinn-With-Two-Ns: this example runs a local activity right?  A local activity modifies workflow state so it is not allowed. The rules for what is allowed in a validator are the same as for a query handler.

tsurdilo: @Quinn-With-Two-Ns updated the code to not invoke LA from update validator. let me know if ok now

tsurdilo: just wanted to add, having to throw TemporalFailure from update handler is weird. If this is done in workflow code the wf execution would fail, same if you did in signal handlers, so thats why it just feels weird that you have to do it and it does not fail execution just the update

bergundy: We'll likely not support local activities in validators but there has been talk of adding a different concept to SDKs which is backed by activities but doesn't generate history events that can be used in both query handlers and update validators.

bergundy: We should throw in Java (and Go) if any commands are being generated in query handlers and update validators (and side-effects FTR).

Quinn-With-Two-Ns: >ok. should we then open issue for it in java sdk to throw error and not allow temporal apis?
Yes, [we already do ](https://github.com/temporalio/sdk-java/issues/1748)

>  just wanted to add, having to throw TemporalFailure from update handler is weird
Do you have a suggestion on how you would implement it differently? I think the issue is your thinking of the update handler as part of the workflow like signal, but really its another sub request running inside a workflow that can succeed/fail on its own.

springboot/src/main/java/io/temporal/samples/springboot/SamplesController.java
```diff
@@ -64,4 +72,52 @@ public String metrics(Model model) {
...
+            PurchaseWorkflow.class,
+            WorkflowOptions.newBuilder()
+                .setTaskQueue("UpdateSampleTaskQueue")
+                .setWorkflowId("NewPurchase")
+                .build());
+    WorkflowClient.start(workflow::start);
+
+    // send update
+    try {
+      workflow.makePurchase(purchase);
```
Quinn-With-Two-Ns, line 105: Why don't you make use of the return value `makePurchase`?

springboot/src/main/java/io/temporal/samples/springboot/update/README.md
```diff
@@ -0,0 +1,34 @@
...
+Next pick the amount of this item you want to purchase. 
+The inventory is presented in the table below the form.
+For each item you can see the current availble stock count.
+Try first picking an item and then an amount that is less or equal to the items in 
+inventory. You will see that the purchase goes through and the inventory table is updated
+dynamically.
+
+Now try to pick and item and amount that is greater than whats in our inventory.
+You will see that the update fails and you see the message (IllegalArgumentException)
+with message. This is thrown from our update validator that blocks the update.
```
Quinn-With-Two-Ns, line 21: This is out of date now I believe

tsurdilo: updated

springboot/src/main/java/io/temporal/samples/springboot/SamplesController.java
tsurdilo: updated. nice catch

springboot/src/main/java/io/temporal/samples/springboot/update/PurchaseWorkflowImpl.java
tsurdilo: I think if its possible to have a specific error idk like RejectUpdateException or something like that would be easier for users to understand (and if you throw TemporalFailure for it to still fail workflow if user really wants to do that for some reason).

tsurdilo:
@Quinn-With-Two-Ns made all the updates let me know if its ok to get this in now (sorry im asking just have a Kafka sample coming in and wanted to get this merged fist if possible :) )

Quinn-With-Two-Ns (approved):

springboot/src/main/java/io/temporal/samples/springboot/update/PurchaseWorkflowImpl.java
Quinn-With-Two-Ns: So that doesn't really work because if you ran an activity in a update handle and it failed by default if would fail your whole workflow instead of just your update request.

PR #496, opened 2023-07-01
Fix errorprone and compiler warnings

tsurdilo:
fixes all errorprone and compiler warnings

fixes main readme


cretz:
> fixes all errorprone and compiler warnings

Can you configure CI to fail if any of these come back?

tsurdilo:
I really don't know
Need to see if its possible with google errorprone plugin

tsurdilo:
@cretz updated. now build will fail on any issues

Quinn-With-Two-Ns (approved):

PR #516, opened 2023-08-23
Add sample for typed search attributes

Quinn-With-Two-Ns:
Add sample for the new typed search attributes API

### DIFF
--- a/core/src/main/java/io/temporal/samples/hello/README.md
+++ b/core/src/main/java/io/temporal/samples/hello/README.md
@@ -28,6 +28,7 @@ To run each hello world sample, use one of the following commands:
 ./gradlew -q execute -PmainClass=io.temporal.samples.hello.HelloSchedules
 ./gradlew -q execute -PmainClass=io.temporal.samples.hello.HelloSignal
 ./gradlew -q execute -PmainClass=io.temporal.samples.hello.HelloSearchAttributes
+./gradlew -q execute -PmainClass=io.temporal.samples.hello.HelloTypedSearchAttributes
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/temporal/samples/hello/README.md
```diff
@@ -28,6 +28,7 @@ To run each hello world sample, use one of the following commands:
 ./gradlew -q execute -PmainClass=io.temporal.samples.hello.HelloSchedules
 ./gradlew -q execute -PmainClass=io.temporal.samples.hello.HelloSignal
 ./gradlew -q execute -PmainClass=io.temporal.samples.hello.HelloSearchAttributes
```
cretz, line 30: I wonder if we should toss a deprecation notice in on this one

cretz (approved):

core/src/main/java/io/temporal/samples/hello/README.md
Quinn-With-Two-Ns: The API is marked deprecated, that seems enough for me?

cretz: Can't see that when browsing samples. I'll just go to https://github.com/temporalio/samples-java/tree/main/core/src/main/java/io/temporal/samples/hello looking how to do search attributes, read https://github.com/temporalio/samples-java/blob/main/core/src/main/java/io/temporal/samples/hello/HelloSearchAttributes.java, and assume it's a suggested sample/approach. Just a note on sample code itself I think has value. But not blocking this issue.

PR #529, opened 2023-09-28
Update HelloSearchAttribute sample to use typed SA api

tsurdilo:
updates hello SA sample to use typed api

### DIFF
--- a/core/src/main/java/io/temporal/samples/hello/HelloSearchAttributes.java
+++ b/core/src/main/java/io/temporal/samples/hello/HelloSearchAttributes.java
@@ -23,12 +23,14 @@ import io.temporal.activity.ActivityInterface;
 import io.temporal.activity.ActivityMethod;
 import io.temporal.activity.ActivityOptions;
 import io.temporal.api.common.v1.Payload;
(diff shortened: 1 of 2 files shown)


Quinn-With-Two-Ns:
Is this sample not sufficient? https://github.com/temporalio/samples-java/blob/main/core/src/main/java/io/temporal/samples/hello/HelloTypedSearchAttributes.java

tsurdilo:
Yes only thing is that the HelloSearchAttribute one had deprecated api calls, maybe worth still changing. Wdyt?

tsurdilo:
I think we can remove the old one (one i changed here) but just update the test to use HelloTypedSearchAttributes, wdyt?

Quinn-With-Two-Ns:
@tsurdilo sorry for the late reply, go sick and forgot about this PR, I think it makes sense to mark `HelloSearchAttribute`  as deprecated or delete and rename `HelloTypedSearchAttributes` to `HelloSearchAttribute`

Originally I kept  `HelloSearchAttribute`  because  some users don't upgrade the SDK often so they might not have the new API .

PR #573, opened 2024-01-26
add cancelation sample with workflow timer

tsurdilo:
adds sample that cancels activity and completes exec on workflow timer firing

this can be linked to when we tell users not to use workflow run/execution timeout which is not recommended

### DIFF
--- a/core/src/main/java/io/temporal/samples/hello/HelloCancellationScopeWithTimer.java
+++ b/core/src/main/java/io/temporal/samples/hello/HelloCancellationScopeWithTimer.java
@@ -1,0 +1,231 @@
+/*
+ *  Copyright (c) 2020 Temporal Technologies, Inc. All Rights Reserved
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/temporal/samples/hello/HelloCancellationScopeWithTimer.java
```diff
@@ -0,0 +1,231 @@
...
+
+  // Define the workflow implementation which implements our getGreeting workflow method.
+  public static class CancellationWithTimerWorkflowImpl implements CancellationWithTimerWorkflow {
+    private final UpdateInfoActivities activities =
+        Workflow.newActivityStub(
+            UpdateInfoActivities.class,
+            ActivityOptions.newBuilder()
+                // If heartbeat timeout is not set, activity heartbeats will be throttled to one
+                // every 30 seconds
+                .setHeartbeatTimeout(Duration.ofSeconds(2))
```
Quinn-With-Two-Ns, line 86: The activity also just won't have a heartbeat timeout

tsurdilo: ugh...forgot. fixing

Quinn-With-Two-Ns (approved):

core/src/main/java/io/temporal/samples/hello/HelloCancellationScopeWithTimer.java
tsurdilo: thanks. updated

PR #574, opened 2024-01-29
Add basic SpringBoot sample

Quinn-With-Two-Ns:
Add basic springboot sample with minimal dependencies just showcasing Temporal + Springboot


antmendoza (approved):
thanks @Quinn-With-Two-Ns

tsurdilo (approved):

PR #581, opened 2024-02-19
SpringBoot - Apache Camel Sample

tsurdilo:
This sample shows possible integration of Temporal in Apache Camel application. 
Shows how to define a Camel Route and start a Workflow from it to gain advantage of Temporal.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,6 +29,7 @@ subprojects {
         otelVersion = '1.30.1'
         otelVersionAlpha = "${otelVersion}-alpha"
         javaSDKVersion = '1.22.3'
+        camelVersion = '3.22.1'
         jarVersion = '1.0.0'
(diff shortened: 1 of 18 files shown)


build.gradle
```diff
@@ -29,6 +29,7 @@ subprojects {
         otelVersion = '1.30.1'
         otelVersionAlpha = "${otelVersion}-alpha"
         javaSDKVersion = '1.22.3'
+        camelVersion = '3.22.1'
```
Quinn-With-Two-Ns, line 32: Any reason you used `3.22.1` vs `4.x`?

tsurdilo: 4.x requires java 17 or higher

Quinn-With-Two-Ns (approved):

PR #593, opened 2024-02-28
Update request processor

dandavison:
[Draft]


core/src/main/java/io/temporal/samples/hello/CommandProcessor.java
```diff
@@ -0,0 +1,137 @@
...
+    @Override
+    public String submitCommand(int command) {
+      if (command < 0) {
+        this.done = true;
+        return "stopping workflow";
+      }
+      CompletablePromise<String> promise = Workflow.newPromise();
+      QueuedCommand queuedCommand = new QueuedCommand(command, promise);
+      this.commandQueue.add(queuedCommand);
+      return queuedCommand.promise.get();
```
mjameswh, line 86: @Quinn-With-Two-Ns Is there a way here to free up the update thread? e.g. could the update handler return the promise, rather than block on `.get()`?

core/src/main/java/io/temporal/samples/hello/CommandProcessor.java
```diff
@@ -0,0 +1,137 @@
...
+      public int command;
+      public CompletablePromise<String> promise;
+
+      public QueuedCommand(int command, CompletablePromise<String> promise) {
+        this.command = command;
+        this.promise = promise;
+      }
+    }
+
+    private ArrayList<QueuedCommand> commandQueue;
```
mjameswh, line 51: `ArrayList` works in this case, but prefer `WorkflowQueue` instead. It provides a more appropriate API, inspired from "Blocking Queue", but Workflow-safe (it internally use `Workflow.await()` to implement blocking ops).

dandavison: Thanks! Yes @Quinn-With-Two-Ns pointed that out to me also.

core/src/main/java/io/temporal/samples/hello/CommandProcessor.java
dandavison: Good point, I'd be interested to know this, but haven't looked into it yet.

Quinn-With-Two-Ns: Yeah we can explore alternative approaches that use less threads. Given we only allow 10 concurrent updates it wasn't prioritized, virtual threads also make this not a big deal.

PR #616, opened 2024-04-25
Adding accumulator pattern sample

joshmsmith:
## What was changed
Adding accumulator pattern sample: HelloAccumulator and associated readme changes

## Why?
This is a pattern that's pretty common, and I wanted to have a sample that I could share about it

## Checklist

1. Closes 
N/A

2. How was this tested:
Tests in HelloAccumulatorTest.Java
Code in main() to demonstrate functionality.

3. Any docs updates needed?
updated the readmes to mention this new sample.

### DIFF
--- a/core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
+++ b/core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2024 Temporal Technologies, Inc. All Rights Reserved
+ *  Copyright (c) 2020 Temporal Technologies, Inc. All Rights Reserved
  *
  *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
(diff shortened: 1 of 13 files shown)


core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
```diff
@@ -0,0 +1,625 @@
...
+        }
+      } while (!unprocessedGreetings.isEmpty() || !Workflow.getInfo().isContinueAsNewSuggested());
+
+      logger.info("starting continue as new processing");
+
+      // Create a workflow stub that will be used to continue this workflow as a new
+      AccumulatorWorkflow continueAsNew = Workflow.newContinueAsNewStub(AccumulatorWorkflow.class);
+
+      // Request that the new run will be invoked by the Temporal system:
+      continueAsNew.accumulateGreetings(bucketKey, greetings, allGreetingsSet);
```
tsurdilo, line 245: you could get buffered signal(s) at this point, guess need to decide to handle them or send them to next exec

joshmsmith: Thanks for the comment!

Is this in the case of a high volume of signals, so that one would come in between the do loop above ending and the continue as new happening?

Do we have sample code that shows how to work with buffered signals?

core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
```diff
@@ -0,0 +1,625 @@
...
+      //    - wait for signals to come in
+      //    - every time a signal comes in, wait again for MAX_AWAIT_TIME
+      //    - if time runs out, and there are no messages, process them all and exit
+      //    - if exit signal is received, process any remaining signals and exit
+      do {
+
+        Workflow.await(MAX_AWAIT_TIME, () -> !unprocessedGreetings.isEmpty() || stopWaiting);
+
+        if (!unprocessedGreetings.isEmpty()) {
+          processGreeting(unprocessedGreetings.removeFirst());
```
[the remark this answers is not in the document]
joshmsmith: It's a good question. Do you mean, many process more than one greeting instead of just one here per loop? Or am I missing what you mean by "outside this loop"?

antmendoza (commented):
Thanks @joshmsmith

core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
```diff
@@ -0,0 +1,625 @@
...
+  public static class AccumulatorWorkflowImpl implements AccumulatorWorkflow {
+
+    private final GreetingActivities activities =
+        Workflow.newActivityStub(
+            GreetingActivities.class,
+            ActivityOptions.newBuilder().setStartToCloseTimeout(Duration.ofSeconds(2)).build());
+
+    private static final Logger logger = LoggerFactory.getLogger(AccumulatorWorkflowImpl.class);
+    String bucketKey;
+    ArrayDeque<Greeting> greetings;
```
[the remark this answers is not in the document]
joshmsmith: I think doing this will break the tests in HelloAccumulatorTest. Most of the other samples have their Impl as public. 
Setting Greeting as private also breaks the test.

core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
```diff
@@ -0,0 +1,625 @@
...
+    HashSet<String> allGreetingsSet;
+    boolean stopWaiting = false;
+    ArrayDeque<Greeting> unprocessedGreetings;
+
+    @Override
+    public String accumulateGreetings(
+        String bucketKeyInput,
+        ArrayDeque<Greeting> greetingsInput,
```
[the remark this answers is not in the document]
joshmsmith: this is good, thank you.

core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
```diff
@@ -0,0 +1,625 @@
...
+        return;
+      }
+
+      if (!greeting.bucket.equals(bucketKey)) {
+        logger.warn("wrong bucket, something is wrong with your signal processing: " + greeting);
+        return;
+      }
+
+      if (allGreetingsSet == null) {
+        logger.warn("allGreetingsSet is null - " + allGreetingsSet);
```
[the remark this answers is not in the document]
joshmsmith: good call, i meant to remove it. Removing.

core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
```diff
@@ -0,0 +1,625 @@
...
+        if (!unprocessedGreetings.isEmpty()) {
+          processGreeting(unprocessedGreetings.removeFirst());
+        } else {
+          String greetEveryone = processGreetings(greetings);
+
+          if (unprocessedGreetings.isEmpty()) {
+            logger.info("Greeting queue is still empty");
+            return greetEveryone;
+          } else if (stopWaiting && unprocessedGreetings.isEmpty()) {
+            return greetEveryone;
```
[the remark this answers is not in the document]
joshmsmith: This is a good catch, taking it out. It was leftover from some testing I was doing with signals at end of workflow

core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
```diff
@@ -0,0 +1,625 @@
...
+        logger.info("Duplicate signal event: " + greeting.greetingKey);
+        return;
+      }
+
+      // add in any desired event processing activity here
+      greetings.add(greeting);
+    }
+
+    private String processGreetings(ArrayDeque<Greeting> greetings) {
+      logger.info("Timer fired for WorkflowId: " + Workflow.getInfo().getWorkflowId());
```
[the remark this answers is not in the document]
joshmsmith: Good point, I will clarify it.

core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
```diff
@@ -0,0 +1,625 @@
...
+    }
+
+    // Test sending more signals after workflow exit
+    try {
+      // send a second workflow signal
+      Greeting janeGreeting = new Greeting("Jane Robot", bucket, "112358132134");
+      workflowSync.waitforGreeting(janeGreeting);
+
+      if (testBadBucket) {
+        // send a third signal with an incorrect bucket
```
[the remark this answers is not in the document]
joshmsmith: This is an excellent point. I will mention it.

core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
drewhoskins-temporal: We could likely make an improvement to this by using our upcoming are_handlers_finished API, once it's ready.

core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
```diff
@@ -0,0 +1,619 @@
...
+  public interface AccumulatorWorkflow {
+    /**
+     * This is the method that is executed when the Workflow Execution is started. The Workflow
+     * Execution completes when this method finishes execution.
+     */
+    @WorkflowMethod
+    String accumulateGreetings(
+        String bucketKey, Deque<Greeting> greetings, Set<String> allGreetingsSet);
+
+    // Define the workflow waitForName signal method. This method is executed when
```
drewhoskins-temporal, line 144: Do you need the first sentence at all?  If so, I think `waitForName` is an older name?

core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
```diff
@@ -0,0 +1,619 @@
...
+    // the workflow receives a greeting signal.
+    @SignalMethod
+    void waitforGreeting(Greeting greeting);
```
drewhoskins-temporal, line 147: This name confused me.  From the client's perspective, they're calling this signal, but they're not waiting for anything.  Typically, handlers are just named after the signal that's being sent, i.e. `sendGreeting`.

core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
drewhoskins-temporal: @joshmsmith as I understand it, you're doing it correctly.  The server will fail the workflow task if any signals come in after this, and the task will be retried.

Aside: We're adding a predicate called something like `allMessagesHandled` that you'll be able to check, but it's more for people that aren't already enqueueing requests.

core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
```diff
@@ -0,0 +1,619 @@
...
+      // ../../updatabletimer
+
+      // Main Workflow Loop:
+      //    - wait for signals to come in
+      //    - every time a signal comes in, wait again for MAX_AWAIT_TIME
+      //    - if time runs out, and there are no messages, process them all and exit
+      //    - if exit signal is received, process any remaining signals and exit
+      do {
+
+        Workflow.await(MAX_AWAIT_TIME, () -> !unprocessedGreetings.isEmpty() || stopWaiting);
```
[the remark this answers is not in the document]
joshmsmith: I updated this, thanks.

core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
joshmsmith: good call, changed.

core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
joshmsmith: good call, changed.

core/src/main/java/io/temporal/samples/hello/HelloAccumulator.java
```diff
@@ -0,0 +1,634 @@
...
+     ArrayDeque<Greeting> greetings;
+     HashSet<String> allGreetingsSet;
+     boolean exitRequested = false;
+     ArrayDeque<Greeting> unprocessedGreetings;
+ 
+     @Override
+     public String accumulateGreetings(
+         String bucketKeyInput, Deque<Greeting> greetingsInput, Set<String> allGreetingsSetInput) {
+       bucketKey = bucketKeyInput;
+       if (unprocessedGreetings == null) unprocessedGreetings = new ArrayDeque<Greeting>();
```
Quinn-With-Two-Ns, line 201: nit pick: you don't need this check you can initialize `unprocessedGreetings` on line 195 and avoid this check and the same check in the signal handler

Quinn-With-Two-Ns: You can see use do it in this sample https://github.com/temporalio/samples-java/blob/main/core/src/main/java/io/temporal/samples/moneybatch/AccountTransferWorkflowImpl.java#L43

Quinn-With-Two-Ns (approved):

PR #648, opened 2024-07-13
update dsl sample, make it simpler, more domain specific

tsurdilo:
dsl sample rewrite to make it much simpler and more intuitive


README.md
```diff
@@ -89,7 +89,7 @@ See the README.md file in each main sample directory for cut/paste Gradle comman
 
 - [**Money Batch**](/core/src/main/java/io/temporal/samples/moneybatch): Demonstrates a situation where a single deposit should be initiated for multiple withdrawals. For example, a seller might want to be paid once per fixed number of transactions. This sample can be easily extended to perform a payment based on more complex criteria, such as at a specific time or an accumulated amount. The sample also demonstrates how to Signal the Workflow when it executes (*Signal with start*). If the Workflow is already executing, it just receives the Signal. If it is not executing, then the Workflow executes first, and then the Signal is delivered to it. *Signal with start* is a "lazy" way to execute Workflows when Signaling them.
 
-- [**Customer Application Approval DSL**](/core/src/main/java/io/temporal/samples/dsl): Demonstrates execution of a customer application approval workflow defined in a DSL (like JSON or YAML)
+- [**DSL (JSON)**](/core/src/main/java/io/temporal/samples/dsl): Demonstrates using domain specific language (DSL) defined in JSON to specify sequence of steps to be performed in our workflow.
```
Quinn-With-Two-Ns, line 92: Nit: could we use the full title here. This is such a nitpick, but I do think for someone skimming through the readme it would be better.

Quinn-With-Two-Ns:
For my own understanding, did we receive feedback from users that the current sample was to confusing?

README.md
tsurdilo: done. thanks.

tsurdilo:
@Quinn-With-Two-Ns yes that was one thing, another was it was using a generic dsl library and typically users are looking at implementing the json/yaml themselves to make it very specific to their needs

jalotra:
hey @tsurdilo is there any roadmap on how the DSL feature is going to pan out in coming months ? 
I want to understand 2 things : 
a) what are the lacking features here, that devs should have to use code to solve along with writing DSL ? 
b) what are some of the other people / tech companies using DSL a lot right now to define their workflows ?

javier-carvajal:
Hello @tsurdilo,

Thank you for your work!

I have a couple of questions:
1-  Is Temporal ever going to support Serverless Workflow?
2- Do you know of a DSL 'interpreter' project for the Java SDK that is versatile enough to build dynamic and complex use cases (similar to BPM)?

Thanks in advance for your help!

bloody0748:
@tsurdilo 

Could you please let me know if the DSL examples for temporal-java are compliant with the latest Serverless Workflow Specification 1.00?

 Are there any plans to update with more examples in the future, such as switch, event, and delay? 
 
 Additionally, I would like to ask if the Serverless Workflow Java SDK will be updated to comply with Specification 1.00.

Thank you for your assistance.

PR #649, opened 2024-07-15
Add polling sample for infrequent poll with Retry-After delay - showc…

tsurdilo:
this adds a polling sample that shows use of ApplicationFailure.newFailureWithCauseAndDelay
and how it can be used when downstream service activity invokes sends back Retry-After header in response 
that defines next retry of polling activity


tsurdilo:
@cretz hi, thanks for review, yes will work on it.

PR #664, opened 2024-08-18
sample that shows how to set custom change version search attribute w…

tsurdilo:
sample that shows how to upsert custom change version search attribute until automatic upsert of TemporalChangeVersion is implemented in sdk
adding so it can be linked to users that require this feature currently

### DIFF
--- a/core/src/main/java/io/temporal/samples/customchangeversion/CustomChangeVersionWorkflowImpl.java
+++ b/core/src/main/java/io/temporal/samples/customchangeversion/CustomChangeVersionWorkflowImpl.java
@@ -1,0 +1,72 @@
+/*
+ *  Copyright (c) 2020 Temporal Technologies, Inc. All Rights Reserved
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/temporal/samples/customchangeversion/CustomChangeVersionWorkflowImpl.java
```diff
@@ -0,0 +1,72 @@
...
+  public String run(String input) {
+    String result = activities.customOne(input);
+    // We assume when this change is added we have some executions of this workflow type running
+    // where customTwo activity was not called (we are adding it to exiting workflow)
+    // Adding customTwo activity as a versioned change
+    int version = Workflow.getVersion("add-v2-activity-change", Workflow.DEFAULT_VERSION, 1);
+    if (version == 1) {
+      // Upsert our custom change version search attribute
+      // We set its value to follow TemporalChangeVersion structure of "<changeId>-<version>"
+      Workflow.upsertTypedSearchAttributes(
```
Quinn-With-Two-Ns, line 52: You should probably use a different SA for each version no?

tsurdilo: may i ask why? one execution moves to new version do users need to know that it was on previous one before? not sure if TemporalChangeVersion has that capability

Quinn-With-Two-Ns: Sorry maybe I wasn't clear by "each version" I meant each `changeId`. `TemporalChangeVersion` is actually a keyword list in other SDKs.

Quinn-With-Two-Ns (approved):

PR #673, opened 2024-09-09
Adding update handler to dynamic sample

tsurdilo:
adds dynamic update handler and validator to DynamicSample

### DIFF
--- a/core/src/main/java/io/temporal/samples/hello/HelloDynamic.java
+++ b/core/src/main/java/io/temporal/samples/hello/HelloDynamic.java
@@ -19,21 +19,27 @@
 
 package io.temporal.samples.hello;
 
+import com.google.protobuf.InvalidProtocolBufferException;
 import io.temporal.activity.Activity;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/temporal/samples/hello/HelloDynamic.java
```diff
@@ -54,18 +60,57 @@ public Object execute(EncodedValues args) {
       // Register dynamic signal handler
       Workflow.registerListener(
           (DynamicSignalHandler)
-              (signalName, encodedArgs) -> name = encodedArgs.get(0, String.class));
-
+              (signalName, encodedArgs) -> names.add(encodedArgs.get(0, String.class)));
+
+      // Register dynamic update handler
+      DynamicUpdateHandler updateHandler =
```
Quinn-With-Two-Ns, line 66: I think the `DynamicUpdateHandler ` API here is flawed, we should change the API before adding a sample

PR #681, opened 2024-10-04
safe message passing sample

Quinn-With-Two-Ns:
Add safe message passing sample. 

Code based on https://github.com/temporalio/samples-python/tree/main/message_passing/safe_message_handlers

### DIFF
--- a/core/src/main/java/io/temporal/samples/safemessagepassing/ClusterManagerActivitiesImpl.java
+++ b/core/src/main/java/io/temporal/samples/safemessagepassing/ClusterManagerActivitiesImpl.java
@@ -69,4 +69,14 @@ public class ClusterManagerActivitiesImpl implements ClusterManagerActivities {
     }
     return badNodes;
   }
+
+  @Override
+  public void shutdown() {
+    log.info("Shutting down cluster");
+    try {
+      Thread.sleep(100);
+    } catch (InterruptedException e) {
+      throw new RuntimeException(e);
+    }
+  }
 }
--- a/core/src/main/java/io/temporal/samples/safemessagepassing/ClusterManagerActivitiesImpl.java
+++ b/core/src/main/java/io/temporal/samples/safemessagepassing/ClusterManagerActivitiesImpl.java
@@ -1,3 +1,22 @@
+/*
+ *  Copyright (c) 2020 Temporal Technologies, Inc. All Rights Reserved
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/temporal/samples/safemessagepassing/ClusterManagerWorkflow.java
```diff
@@ -0,0 +1,169 @@
...
+  // This is an update as opposed to a signal because the client may want to wait for nodes to be
+  // allocated
+  // before sending work to those nodes.
+  // Returns the list of node names that were allocated to the job.
+  @UpdateMethod
+  ClusterManagerAssignNodesToJobResult assignNodesToJobs(ClusterManagerAssignNodesToJobInput input);
+
+  // Even though it returns nothing, this is an update because the client may want to track it, for
+  // example
+  // to wait for nodes to be unassigned before reassigning them.
```
dandavison, line 166:
```suggestion
  // Even though it returns nothing, this is an update because the client may want to track it, for
  // example to wait for nodes to be unassigned before reassigning them.
```

dandavison (commented):
Nice, all looks good, except I have a concern that we're doing start/stop wrong in all our samples (except Go?).

core/src/main/java/io/temporal/samples/safemessagepassing/ClusterManagerActivitiesImpl.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @Override
+  public void assignNodesToJob(AssignNodesToJobInput input) {
+    for (String node : input.getNodes()) {
+      log.info("Assigned node " + node + " to job " + input.getJobName());
+    }
+    try {
+      Thread.sleep(100);
+    } catch (InterruptedException e) {
+      throw new RuntimeException(e);
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: `InterruptedException ` is a checked exception so we need to translate it to a unchecked exception

Quinn-With-Two-Ns:
@dandavison thanks for the feedback, I updated the sample based on what you did in TS

core/src/main/java/io/temporal/samples/safemessagepassing/ClusterManagerWorkflowImpl.java
```diff
@@ -0,0 +1,244 @@
...
+    for (int i = 0; i < 25; i++) {
+      state.nodes.put(String.valueOf(i), Optional.empty());
+    }
+    logger.info("Cluster started");
+  }
+
+  @Override
+  public boolean stopCluster() {
+    if (state.workflowState != ClusterState.STARTED) {
+      // This is used as an Update handler we return an error to the caller.
```
dandavison, line 114:
```suggestion
      // This is used as an Update handler so that we can return an error to the caller.
```

dandavison (approved):
LGTM!

PR #686, opened 2024-10-15
Add Nexus sample

Quinn-With-Two-Ns:
Add Nexus sample based on the Go SDK sample https://github.com/temporalio/samples-go/tree/main/nexus


bergundy (approved):
Skimmed this, seems like a fairly straightforward translation.

PR #688, opened 2024-10-18
Fix nexus sample TLS settings

Quinn-With-Two-Ns:
Fix nexus sample TLS settings to be the same as other TLS samples and update the `service.md`


bergundy (approved):
Rubber stamp.

cretz (approved):

PR #689, opened 2024-10-18
EarlyReturn sample (Update-With-Start)

steveandroulakis:
If it is a significant code change, please **make sure there is an open issue** for this. 
We work best with you when we have accepted the idea first before you code. -->

## What was changed
Sample that uses the new Update-With-Start feature.

## Why?
Help users understand what Update-With-Start can do with a simple financial transaction example.

Based on the Go sample here: https://github.com/temporalio/samples-go/tree/main/early-return

1. Closes
No issue

2. How was this tested:
I welcome guidance on how to effectively write tests for this!

3. Any docs updates needed?
No


mfateev:
We should either use the proper Java style of separating classes and interfaces into their own files or implement this as another Hello sample.

core/src/main/java/io/temporal/samples/earlyreturn/EarlyReturn.java
```diff
@@ -0,0 +1,256 @@
...
+            .build();
+
+    WorkflowStub workflowStub = client.newUntypedWorkflowStub("TransactionWorkflow", options);
+
+    try {
+      System.out.println("Starting workflow with UpdateWithStart");
+
+      UpdateWithStartWorkflowOperation<String> update =
+          UpdateWithStartWorkflowOperation.newBuilder(UPDATE_NAME, String.class, new Object[] {})
+              .setWaitForStage(WorkflowUpdateStage.COMPLETED)
```
Quinn-With-Two-Ns, line 81: Is there a reason you chose to use the untyped API here?

core/src/main/java/io/temporal/samples/earlyreturn/EarlyReturn.java
```diff
@@ -0,0 +1,256 @@
...
+    private void sleep(long millis) {
+      try {
+        Thread.sleep(millis);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      }
+    }
+  }
+
+  public static class Transaction {
```
Quinn-With-Two-Ns, line 209: For jackson serialization we should use annotations so we don't need to expose a bunch of extra method for serialization.

We have one example here https://github.com/temporalio/samples-java/blob/main/core/src/main/java/io/temporal/samples/nexus/service/NexusService.java#L62C5-L62C17

Oracle blog recommend the same approach https://blogs.oracle.com/javamagazine/post/java-json-serialization-jackson

steveandroulakis:
> We should either use the proper Java style of separating classes and interfaces into their own files or implement this as another Hello sample.

Thanks, will split into separate files.

core/src/main/java/io/temporal/samples/earlyreturn/EarlyReturn.java
steveandroulakis: Nope, will change to typed. Thanks

core/src/main/java/io/temporal/samples/earlyreturn/EarlyReturn.java
steveandroulakis: Got it, will implement. Thanks

core/src/main/java/io/temporal/samples/earlyreturn/EarlyReturn.java
steveandroulakis: This change is blocked by https://github.com/temporalio/sdk-java/issues/2285

Will update once it's fixed

steveandroulakis: Tested with commit from fix PR https://github.com/temporalio/sdk-java/pull/2286

core/src/main/java/io/temporal/samples/earlyreturn/TransactionWorkflowImpl.java
```diff
@@ -0,0 +1,68 @@
...
+      activities.cancelTransaction(this.tx);
+      return "Transaction cancelled";
+    } else {
+      activities.completeTransaction(this.tx);
+      return "Transaction completed successfully: " + this.tx.getId();
+    }
+  }
+
+  @Override
+  public String returnInitResult() {
```
drewhoskins-temporal, line 60: nit: We generally try to set a good example and return structures rather than scalars for future compat of being able to add new fields.  I was asked to improve a similar situation when I wrote a python sample, so figured I'd echo the advice here.  FF to ignore since I'm not sure how consistent we really are and understand it's orthogonal.

drewhoskins-temporal (commented):
optional nit

core/src/main/java/io/temporal/samples/earlyreturn/TransactionWorkflowImpl.java
```diff
@@ -0,0 +1,70 @@
...
+          TransactionActivities.class,
+          ActivityOptions.newBuilder().setStartToCloseTimeout(Duration.ofSeconds(30)).build());
+
+  private boolean initDone = false;
+  private Transaction tx;
+  private Exception initError = null;
+
+  @Override
+  public TxResult processTransaction(Transaction txInput) {
+    this.tx = txInput;
```
Quinn-With-Two-Ns, line 41: I would refactor this sample to make `Transaction` immutable and not modify transaction as the workflow progress. I would make two different classes one for the workflow input that doesn't have a transaction ID field and one  (maybe returned by the activity `initTransaction` or created in the workflow) that the `cancelTransaction` and `completeTransaction`

core/src/main/java/io/temporal/samples/earlyreturn/TransactionWorkflowImpl.java
```diff
@@ -0,0 +1,70 @@
...
+      activities.cancelTransaction(this.tx);
+      return new TxResult(this.tx.getId(), "Transaction cancelled.");
+    } else {
+      activities.completeTransaction(this.tx);
+      return new TxResult(this.tx.getId(), "Transaction completed successfully.");
+    }
+  }
+
+  @Override
+  public TxResult returnInitResult() {
```
Quinn-With-Two-Ns, line 60: Maybe use a different type here to show the init result does not need to return the same type as the main workflow? Maybe return just the transaction ID here.

Quinn-With-Two-Ns: Saw Drews comment about returning classes, and I generally agree with it, so feel free to disregard this suggestion (could define a new class, but that seems like overkill)

Quinn-With-Two-Ns:
@stephanos May want to take a look

Quinn-With-Two-Ns:
A unit test showing update with start would be great, @stephanos but a lot of work into the time skipping test server part, but I wouldn't block the PR on it if you don't have time.

core/src/main/java/io/temporal/samples/earlyreturn/TransactionWorkflowImpl.java
steveandroulakis: Done: creation of a TransactionRequest object + the Transaction class is now immutable

core/src/main/java/io/temporal/samples/earlyreturn/TransactionWorkflowImpl.java
steveandroulakis: Done

steveandroulakis:
> A unit test showing update with start would be great, @stephanos but a lot of work into the time skipping test server part, but I wouldn't block the PR on it if you don't have time.

I have written 2 unit tests. I hope they're done correctly!

core/src/main/java/io/temporal/samples/earlyreturn/TransactionActivitiesImpl.java
```diff
@@ -0,0 +1,75 @@
...
+            txRequest.getSourceAccount(),
+            txRequest.getTargetAccount(),
+            txRequest.getAmount());
+
+    sleep(500);
+    return tx;
+  }
+
+  @Override
+  public void cancelTransaction(TransactionRequest txRequest) {
```
Quinn-With-Two-Ns, line 49: I think `cancelTransaction` should also take the `Transaction` class since, in practice, you would need the transaction ID to cancel it right?

steveandroulakis: I'm diverging more and more from the original Go sample (which is fine).

Right now, I think of as initialization as something that results in the 'minting' of a Transaction ID. And a failure of initialization causes an activity failure which triggers a cancellation.

I can't think of a way of having
1. InitTransaction cause an activity failure (fundamental to @stephanos original sample) AND
2. Returning a transaction ID that would mean the creation of a Transaction object I can then use in cancelTransaction

Thinking blank-slate for a moment, and about the purpose of an "Early Return" type sample. @Quinn-With-Two-Ns -- What do you think you would want to see from a scenario where a transaction fails on initialization (e.g. the workflow update with start)? What data should be returned and what compensation should take place after? I don't think I should spend any more hours chewing on this sample without figuring that out first. Thanks!

Quinn-With-Two-Ns: Yeah you raise a great point. I think I would separate the creation of the ID from the `initialization` of the transaction. Treating the transaction ID as an idempotency key for the rest of the activities. That being said I think this sample is really good as is and an improvement over the Go Sample so 
 wouldn't hold up the PR on this

Quinn-With-Two-Ns (approved):

core/src/main/java/io/temporal/samples/earlyreturn/TransactionActivities.java
```diff
@@ -0,0 +1,35 @@
...
+
+package io.temporal.samples.earlyreturn;
+
+import io.temporal.activity.ActivityInterface;
+import io.temporal.activity.ActivityMethod;
+
+@ActivityInterface
+public interface TransactionActivities {
+  @ActivityMethod
+  Transaction initTransaction(TransactionRequest tx);
```
stephanos, line 28: nit
```suggestion
  Transaction initTransaction(TransactionRequest txRequest);
```

core/src/main/java/io/temporal/samples/earlyreturn/TransactionActivitiesImpl.java
```diff
@@ -0,0 +1,75 @@
...
+    // Simulate transaction ID generation
+    String txId = "TXID" + String.format("%010d", (long) (Math.random() * 1_000_000_0000L));
+    Transaction tx =
+        new Transaction(
+            txId,
+            txRequest.getSourceAccount(),
+            txRequest.getTargetAccount(),
+            txRequest.getAmount());
+
+    sleep(500);
```
stephanos, line 44: nit: is the second sleep necessary?

steveandroulakis: It's intentional. I wanted to give the impression of a failed initialization taking less long than a successful one (since the failed one throws an exception 'early').

Since the sleeps are for artistic effect, this is me artistically expressing myself ;) Let me know if you disagree with this approach and I'll do whatever you want

core/src/main/java/io/temporal/samples/earlyreturn/EarlyReturnClient.java
```diff
@@ -0,0 +1,90 @@
...
+    } catch (Exception e) {
+      System.err.println("Transaction initialization failed: " + e.getMessage());
+    }
+  }
+
+  // Build WorkflowOptions with task queue and unique ID
+  private static WorkflowOptions buildWorkflowOptions() {
+    return WorkflowOptions.newBuilder()
+        .setTaskQueue(TASK_QUEUE)
+        .setWorkflowId(WORKFLOW_ID_PREFIX + System.currentTimeMillis())
```
stephanos, line 87: In the Go example, I'm generated the transaction ID on the client and use it in the Workflow ID. That way, the user would receive an error when they start a workflow for the same transaction again. I don't think it's strictly necessary to mimic that here, but I thought I'd mention it.

core/src/main/java/io/temporal/samples/earlyreturn/TransactionActivitiesImpl.java
stephanos: 👍 makes sense!

core/src/main/java/io/temporal/samples/earlyreturn/EarlyReturnClient.java
steveandroulakis: Gotcha. I intentionally changed it so the early-return would contain meaningful data usable for the rest of the transaction (an early-return / initialization step that returns null isn't very illustrative of real-world value imo)

I plan on following Quinn's suggestion here of still having the ID minted as part of an activity, but doing it in a separate one as part of the initialization step https://github.com/temporalio/samples-java/pull/689#discussion_r1817000501 -- Let me know if there's anything you'd do differently

core/src/test/java/io/temporal/samples/earlyreturn/TransactionWorkflowTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+    // Verify the exception chain
+    assertTrue(exception.getCause() instanceof WorkflowUpdateException);
+    assertTrue(exception.getCause().getCause() instanceof ActivityFailure);
+    ApplicationFailure appFailure = (ApplicationFailure) exception.getCause().getCause().getCause();
+    assertEquals("InvalidAmount", appFailure.getType());
+    assertTrue(appFailure.getMessage().contains("Invalid Amount"));
+
+    // Let workflow process error handling
+    testWorkflowRule.getTestEnvironment().sleep(Duration.ofSeconds(1));
```
stephanos, line 145: Out of curiosity, is this sleep necessary? I would have expected the `getResult` further down would just block until the result is available?

steveandroulakis: You're absolutely right. There's no need as getResult blocks for a response. Removing

PR #700, opened 2024-11-03
Add sample showing AWS encryption SDK

Quinn-With-Two-Ns:
Add sample showing how a user could use the AWS encryption SDK to encrypt payloads with AWS KMS

### DIFF
--- a/core/src/main/java/io/temporal/samples/keymanagementencryption/awsencryptionsdk/KeyringCodec.java
+++ b/core/src/main/java/io/temporal/samples/keymanagementencryption/awsencryptionsdk/KeyringCodec.java
@@ -28,6 +28,7 @@ import io.temporal.payload.codec.PayloadCodec;
 import io.temporal.payload.context.ActivitySerializationContext;
 import io.temporal.payload.context.HasWorkflowSerializationContext;
 import io.temporal.payload.context.SerializationContext;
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/temporal/samples/keymanagementencryption/awsencryptionsdk/KeyringCodec.java
```diff
@@ -0,0 +1,155 @@
...
+      activityType = ((ActivitySerializationContext) serializationContext).getActivityType();
+    }
+    String signature = activityType != null ? workflowId + activityType : workflowId;
+    return Collections.singletonMap("signature", signature);
+  }
+
+  private Payload encodePayload(Payload payload) {
+    byte[] plaintext = payload.toByteArray();
+    byte[] ciphertext =
+        crypto.encryptData(kmsKeyring, plaintext, getEncryptionContext()).getResult();
```
cretz, line 134: Is this call potentially slow (i.e. meaning ever makes a network call, even once lazily)? Should the deadlock detector be disabled or is it already disabled?

Quinn-With-Two-Ns: Yeah good point, I added a line to disable the deadlock detector.

cretz (approved):

PR #701, opened 2024-11-04
Fix formatting

Quinn-With-Two-Ns:
Fix formatting check https://github.com/temporalio/samples-java/actions/runs/11668460249/job/32488165277


cretz (approved):

Quinn-With-Two-Ns:
Yeah I think the issue was I updated the formatter version with another PR in flight

PR #708, opened 2024-12-30
Workflow timer sample

tsurdilo:
sample that shows use of workflow timer instead of using workflow run/exec timeouts

### DIFF
--- a/core/src/main/java/io/temporal/samples/hello/HelloWorkflowTimer.java
+++ b/core/src/main/java/io/temporal/samples/hello/HelloWorkflowTimer.java
@@ -1,0 +1,253 @@
+/*
+ *  Copyright (c) 2020 Temporal Technologies, Inc. All Rights Reserved
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/temporal/samples/hello/HelloWorkflowTimer.java
```diff
@@ -0,0 +1,253 @@
...
+                        // https://github.com/temporalio/sdk-java/issues/87
+                        .thenApply(
+                            ignore -> {
+                              // Cancel the workflow cancellation scope allowing us to react to this
+                              // timer firing
+                              if (workflowCancellationScope != null) {
+                                workflowCancellationScope.cancel("Workflow timer fired");
+                              }
+                              return null;
+                            });
```
antmendoza, line 88: I wonder if having this block inside `workflowCancellationScope` would produce the same output, so we might not need two cencellationScopes

antmendoza (approved):
Thanks @tsurdilo

core/src/main/java/io/temporal/samples/hello/HelloWorkflowTimer.java
tsurdilo: i think you could. having separate maybe cleaner as then we dont cancel scope within itself. wdyt?

PR #719, opened 2025-02-28
Add interceptor to propagate headers to Nexus operations

Quinn-With-Two-Ns:
Add interceptor to propagate headers to Nexus operations


mjameswh:
Why do you need this nexus-specific context propagation as a user-land feature? I mean, shouldn't that be part of a more general "context propagation" sample? And doesn't the SDK already provide context propagation interceptors?

This is not blocking, but you may want to clarify the use case for this vs apparent alternatives in the sample's README file.

mjameswh:
Also:
- Don't you need to also add this sample to the repository's root README file and/or other list of samples?
- Copyright notices seem off. Absolutely no mention of Temporal?

mjameswh (approved):

Quinn-With-Two-Ns:
>  I mean, shouldn't that be part of a more general "context propagation" sample?

Nexus context propagation is different from our other "context propagation" samples since Nexus headers are different then normal Temporal header (map[string]string vs map[string]payload).

> And doesn't the SDK already provide context propagation interceptors?

No, at least not as an interceptor. Java has a concept of a `ContextPropagator` that was added before we had interceptors, but the SDK provides no built in implementation and it is very limited. Only propagates context on workflow and activity start. We want to replace it with a context propagation interceptor but it hasn't been prioritized.

PR #721, opened 2025-03-04
test: added a junit 5 test on SleepForDays

HCaupert:
## What was changed
Added an example for SleepForDays workflow with JUnit5

## Why?
There were no example yet


Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
Thanks for the contribution!

PR #738, opened 2025-05-11
Adds apikey example

stwomack:
## What was changed
Added an example for API Key & Temporal Cloud

## Why?
We didn't have one previously

1. How was this tested:
Via the instructions in the README. It's the same as the ssl example (mTls with certs), and thus doesn't have unit tests

2. Any docs updates needed?
No

### DIFF
--- a/core/src/main/java/io/temporal/samples/apikey/ApiKeyWorker.java
+++ b/core/src/main/java/io/temporal/samples/apikey/ApiKeyWorker.java
@@ -1,0 +1,66 @@
+package io.temporal.samples.apikey;
+
+import io.temporal.client.WorkflowClient;
+import io.temporal.client.WorkflowClientOptions;
+import io.temporal.serviceclient.WorkflowServiceStubs;
+import io.temporal.serviceclient.WorkflowServiceStubsOptions;
+import io.temporal.worker.Worker;
+import io.temporal.worker.WorkerFactory;
+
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/temporal/samples/apikey/ApiKeyWorker.java
```diff
@@ -0,0 +1,66 @@
...
+import io.temporal.serviceclient.WorkflowServiceStubs;
+import io.temporal.serviceclient.WorkflowServiceStubsOptions;
+import io.temporal.worker.Worker;
+import io.temporal.worker.WorkerFactory;
+
+public class ApiKeyWorker {
+  static final String TASK_QUEUE = "MyTaskQueue";
+
+  public static void main(String[] args) throws Exception {
+    // For temporal cloud this would likely be ${namespace}.tmprl.cloud:7233
```
Quinn-With-Two-Ns, line 14: For API keys this is not true, the endpoint is like `{region}.api.temporal.io:7233`

stwomack: Fixed in the worker and the starter

Quinn-With-Two-Ns:
This ample should be linked in the main [readme](https://github.com/temporalio/samples-java/blob/main/README.md)

PR #754, opened 2025-09-16
Create worker deployment based versioning sample

Sushisource:
## What was changed
Updating (adding in this case) versioning samples in all langauges

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/core/src/main/java/io/temporal/samples/workerversioning/README.md
+++ b/core/src/main/java/io/temporal/samples/workerversioning/README.md
@@ -9,7 +9,8 @@ The sample creates multiple worker versions (1.0, 1.1, and 2.0) within one deplo
 
 ## Steps to run this sample:
 
-1) Run a [Temporal service](https://github.com/temporalio/samples-java/tree/main/#how-to-use).
+1) Run a [Temporal service](https://github.com/temporalio/samples-java/tree/main/#how-to-use). And
+   ensure that you're using at least Server version 1.28.0 (CLI version 1.4.0).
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/temporal/samples/workerversioning/README.md
```diff
@@ -0,0 +1,25 @@
...
+This sample demonstrates how to use Temporal's Worker Versioning feature to safely deploy updates to workflow and activity code. It shows the difference between auto-upgrading and pinned workflows, and how to manage worker deployments with different build IDs.
+
+The sample creates multiple worker versions (1.0, 1.1, and 2.0) within one deployment and demonstrates:
+- **Auto-upgrading workflows**: Automatically and controllably migrate to newer worker versions
+- **Pinned workflows**: Stay on the original worker version throughout their lifecycle
+- **Compatible vs incompatible changes**: How to make safe updates using `Workflow.getVersion`
+
+## Steps to run this sample:
+
+1) Run a [Temporal service](https://github.com/temporalio/samples-java/tree/main/#how-to-use).
```
Quinn-With-Two-Ns, line 12: What happens if the user is on an old version of Temporal that doesn't support worker versioning or it isn't enabled? Will this sample fail with a clear error they need to enable it?

Sushisource: It's enabled by default at this point, so the main thing is just to make sure they have a recent version. I've added that to the readme.

PR #776, opened 2026-04-08
Sample code for Nexus messaging

Evanthx:
This is sample code to show two ways to send messages (signals, queries, and updates) through Nexus.


core/src/main/java/io/temporal/samples/nexus_messaging/ondemandpattern/service/NexusRemoteGreetingService.java
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * Nexus service definition for the remote-start pattern. Unlike {@link NexusGreetingService}, every
+ * operation includes a {@code workflowId} so the caller controls which workflow instance is
+ * targeted. This also exposes a {@code runFromRemote} operation that starts a new GreetingWorkflow.
+ */
+@Service
+public interface NexusRemoteGreetingService {
+
+  class RunFromRemoteInput {
+    private final String workflowId;
```
Quinn-With-Two-Ns: So part of the reason to wrap a Workflow in a Nexus service is to hide the workflow from the caller, so I wouldn't call this ID `workflowId` the service doesn't need to expose that it using workflows to the caller.

core/src/main/java/io/temporal/samples/nexus_messaging/ondemandpattern/handler/NexusRemoteGreetingServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+    return WorkflowRunOperation.fromWorkflowHandle(
+        (ctx, details, input) -> {
+          logger.info("RunFromRemote was received for workflow {}", input.getWorkflowId());
+          return WorkflowHandle.fromWorkflowMethod(
+              Nexus.getOperationContext()
+                      .getWorkflowClient()
+                      .newWorkflowStub(
+                          GreetingWorkflow.class,
+                          WorkflowOptions.newBuilder()
+                              .setWorkflowId(input.getWorkflowId())
```
Quinn-With-Two-Ns: Even though it make make the sample a bit more complex, I wonder if we should show here the workflow ID doesn't have to directly map to the nexus operation input. The workflow ID can just be derived/computed liked `.setWorkflowId("NexusRemoteGreetingServiceWorkflow-" + input.getId())`

core/src/main/java/io/temporal/samples/nexus_messaging/README.md
```diff
@@ -0,0 +1,138 @@
+This sample shows how to expose a long-running workflow's queries, updates, and signals as Nexus
+operations. The caller interacts only with the Nexus service; the workflow is a private
+implementation detail.
+
+There are **two caller patterns** that share the same handler workflow (`GreetingWorkflow`):
```
Quinn-With-Two-Ns: I think I would split these up into two separate samples since the amount of code here could be overwhelming and make it harder to follow

core/src/main/java/io/temporal/samples/nexus_messaging/handler/GreetingWorkflowImpl.java
```diff
@@ -0,0 +1,110 @@
...
+    logger.info("validateSetLanguage called");
+    if (!greetings.containsKey(input.getLanguage())) {
+      throw new IllegalArgumentException(input.getLanguage().name() + " is not supported");
+    }
+  }
+
+  @Override
+  public Language setLanguageUsingActivity(NexusGreetingService.SetLanguageInput input) {
+    if (!greetings.containsKey(input.getLanguage())) {
+      // Use a lock so that if this handler is called concurrently, each call executes its activity
```
Quinn-With-Two-Ns: Not sure why we need a lock here? greeting is a map so order shouldn't matter

core/src/main/java/io/temporal/samples/nexus_messaging/caller_remote/CallerRemoteWorkflowImpl.java
```diff
@@ -0,0 +1,78 @@
...
+import io.temporal.workflow.Workflow;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CallerRemoteWorkflowImpl implements CallerRemoteWorkflow {
+
+  private static final Logger logger = LoggerFactory.getLogger(CallerRemoteWorkflowImpl.class);
```
Quinn-With-Two-Ns:
```suggestion
  private static final Logger logger = Workflow.getLogger(CallerRemoteWorkflowImpl.class);
```

Use `Workflow.getLogger` inside workflows

Evanthx: Done

core/src/main/java/io/temporal/samples/nexus_messaging/handler/GreetingWorkflowImpl.java
Evanthx: Done!

core/src/main/java/io/temporal/samples/nexus_messaging/callerpattern/caller/CallerStarter.java
```diff
@@ -0,0 +1,30 @@
...
+import io.temporal.client.WorkflowClientOptions;
+import io.temporal.client.WorkflowOptions;
+import io.temporal.serviceclient.WorkflowServiceStubs;
+import java.util.List;
+import java.util.UUID;
+
+public class CallerStarter {
+
+  public static void main(String[] args) {
+    WorkflowServiceStubs service = WorkflowServiceStubs.newLocalServiceStubs();
```
Quinn-With-Two-Ns: All samples should support env config. You can look at other samples to see how to create the client and follow the pattern there https://github.com/temporalio/samples-java/blob/main/core/src/main/java/io/temporal/samples/asyncchild/Starter.java#L22

core/src/main/java/io/temporal/samples/nexus_messaging/README.md
Evanthx: I agree. Done

core/src/main/java/io/temporal/samples/nexus_messaging/ondemandpattern/handler/NexusRemoteGreetingServiceImpl.java
Evanthx: Good idea, and I learned from executing it. Done.

core/src/main/java/io/temporal/samples/nexus_messaging/callerpattern/caller/CallerStarter.java
Evanthx: Thank you as I didn't know about this - but these seem to be all non-nexus examples? I looked in the other Nexus samples and they aren't using the env files - but they are passing the namespace in on the commandline instead of in code. In code seems clearer to me, but I can switch to commandline arguments if desired.

Quinn-With-Two-Ns: So a bit of background, the reason why all the other samples and the other Nexus samples have this customization point is so users can easily run the samples against a local server and Temporal Cloud. The other Nexus samples simply predate the environment config approach, for this new samples lets use environment config approach since it is the standard.

Quinn-With-Two-Ns:
Java code looks good, Can you share this with the SDK team and the rest of the Nexus team?

core/src/main/java/io/temporal/samples/nexus_messaging/ondemandpattern/service/NexusRemoteGreetingService.java
Evanthx: Done

core/src/main/java/io/temporal/samples/nexus_messaging/callerpattern/caller/CallerStarter.java
Evanthx: Done

core/src/main/java/io/temporal/samples/nexusmessaging/callerpattern/README.md
```diff
@@ -0,0 +1,65 @@
+## Entity pattern
```
cconstable: Is Entity pattern == Caller pattern?

Evanthx: Gah! I thought I'd changed all those!

README.md
```diff
@@ -171,6 +171,9 @@ Load client configuration from TOML files with programmatic overrides.
 
 - [**Context/Header Propagation**](/core/src/main/java/io/temporal/samples/nexuscontextpropagation): Demonstrates how to propagate context through Nexus operation headers.
 
+- [**Nexus Messaging**](/core/src/main/java/io/temporal/samples/nexusmessaging): Demonstrates how send signal, update and query messages through Nexus.
```
tconley1428: How _to_ send?

core/src/main/java/io/temporal/samples/nexusmessaging/README.md
```diff
@@ -0,0 +1,16 @@
+This sample shows how to expose a long-running Workflow's queries, updates, and signals as Nexus
+operations. There are two self-contained examples, each in its own directory:
+
+| | `callerpattern/` | `ondemandpattern/`                                           |
```
tconley1428, line 4: I'm not sure "caller pattern" has a whole lot to imply the workflow is running, but feel free to ignore.

tconley1428 (approved):

README.md
Evanthx: gah! Fixed

PR #778, opened 2026-04-28
Samples for Standalone Activities

GregoryTravis:
Samples for Standalone Activities

Includes both `HelloStandaloneActivity `and a full set of examples in `core/src/main/java/io/temporal/samples/standaloneactivities`.

Tests for both.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,12 +26,13 @@ subprojects {
     ext {
         otelVersion = '1.30.1'
         otelVersionAlpha = "${otelVersion}-alpha"
-        javaSDKVersion = '1.34.0'
+        javaSDKVersion = '1.35.0-SNAPSHOT'
         camelVersion = '3.22.1'
         jarVersion = '1.0.0'
     }
 
     repositories {
+        mavenLocal()
         maven {
             url "https://oss.sonatype.org/content/repositories/snapshots/"
         }
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ subprojects {
     ext {
         otelVersion = '1.30.1'
         otelVersionAlpha = "${otelVersion}-alpha"
-        javaSDKVersion = '1.35.0-SNAPSHOT'
+        javaSDKVersion = '1.35.0'
         camelVersion = '3.22.1'
         jarVersion = '1.0.0'
     }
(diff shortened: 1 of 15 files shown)


build.gradle
```diff
@@ -26,12 +26,13 @@ subprojects {
     ext {
         otelVersion = '1.30.1'
         otelVersionAlpha = "${otelVersion}-alpha"
-        javaSDKVersion = '1.34.0'
+        javaSDKVersion = '1.35.0'
```
GregoryTravis, line 29: Is this correct?

build.gradle
```diff
@@ -26,12 +26,13 @@ subprojects {
...
         camelVersion = '3.22.1'
         jarVersion = '1.0.0'
     }
 
     repositories {
+        mavenLocal()
```
chris-olszewski, line 35: Do we want this?

build.gradle
chris-olszewski: Based on https://github.com/temporalio/samples-java/pull/773, I believe so.

core/src/test/java/io/temporal/samples/standaloneactivities/StandaloneActivitiesTest.java
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+
+/**
+ * Unit tests for the standaloneactivities sample. Uses an embedded Temporal test server so no
+ * external service is required.
+ *
+ * <p>Standalone Activities do not use a Workflow at runtime, but the embedded test server only
+ * supports Activity execution through a Workflow. These tests therefore drive the Activity through
+ * a minimal wrapper Workflow so the Activity logic is exercised against the real SDK worker stack.
```
chris-olszewski, line 25: Not exactly sure what these tests are actually exercising since they aren't leveraging SAA. Just that the activity implementation is sound?

chris-olszewski (approved):
Nothing blocking, unsure about usefulness of tests here, but no harm in including them.

core/src/test/java/io/temporal/samples/standaloneactivities/StandaloneActivitiesTest.java
GregoryTravis: Yeah, these are just a basic smoke test to make sure that the samples don't decay when changes are made in the other repo.

build.gradle
GregoryTravis: Indeed we don't.

PR #782, opened 2026-06-08
Sample code for Nexus Standalone

Evanthx:
Sample code for Standalone Nexus operations. This cannot be checked in before the matching SDK changes - https://github.com/temporalio/sdk-java/pull/2872/


Quinn-With-Two-Ns:
Make sure to link this in the README https://github.com/temporalio/samples-java/blob/main/README.md

core/src/main/java/io/temporal/samples/nexusstandalone/README.MD
```diff
@@ -0,0 +1,99 @@
...
+   (`executeAsync`).
+2. **Cancel** a running operation (`handle.cancel`).
+3. **Terminate** a running operation (`handle.terminate`). Operation-terminate is a known gap that
+   does not stop the backing workflow, so the sample also terminates the backing workflow by ID.
+4. **Visibility** — `list` operations with a status filter and `count` them (total and grouped) via
+   `NexusClient`.
+5. **Client options and interceptors** — set the identity and data converter, and register two
+   logging interceptors.
+
+> [!WARNING]
```
Quinn-With-Two-Ns: NIT: I would put this at the top

core/src/main/java/io/temporal/samples/nexusstandalone/README.MD
```diff
@@ -0,0 +1,99 @@
...
+> Standalone Nexus operations are experimental and may be subject to backwards-incompatible
+> changes. They require a Temporal server that implements and enables them via the dynamic configs
+> shown below.
+
+### Running
+
+Start a Temporal server with the standalone-Nexus dynamic configs enabled:
```
Quinn-With-Two-Ns: We should call out the version of the dev server users need

core/src/main/java/io/temporal/samples/nexusstandalone/service/GreetingIds.java
```diff
@@ -0,0 +1,10 @@
+package io.temporal.samples.nexusstandalone.service;
+
+// A helper method to generate workflow IDs.
+public final class GreetingIds {
```
Quinn-With-Two-Ns: We don't need this helper, just inline this

core/src/main/java/io/temporal/samples/nexusstandalone/service/ClientOptions.java
```diff
@@ -0,0 +1,35 @@
...
+import io.temporal.serviceclient.WorkflowServiceStubs;
+import java.nio.file.Paths;
+
+/**
+ * Builds a {@link WorkflowClient} from the {@code default} profile in {@code
+ * core/src/main/resources/config.toml}. Edit that profile (or override via {@code TEMPORAL_*}
+ * environment variables) to point at a different server or namespace — for example a Temporal Cloud
+ * namespace with an API key.
+ */
+public class ClientOptions {
```
Quinn-With-Two-Ns, line 14: I'd just inline this like all the other samples

core/src/main/java/io/temporal/samples/nexusstandalone/service/ClientOptions.java
```diff
@@ -0,0 +1,35 @@
...
+
+  public static WorkflowClient getWorkflowClient() {
+    ClientConfigProfile profile;
+    try {
+      String configFilePath =
```
Quinn-With-Two-Ns: Why do we have special client config loading code? Other samples just do
```
   ClientConfigProfile profile;
    try {
      profile = ClientConfigProfile.load();
    } catch (IOException e) {
      throw new RuntimeException("Failed to load client configuration", e);
    }
```

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
```diff
@@ -0,0 +1,293 @@
...
+    String namespace = client.getOptions().getNamespace();
+
+    // A single NexusClient is the entry point: it serves Visibility queries (list/count) and
+    // produces service-bound clients.
+    NexusClient nexusClient = NexusClient.newInstance(stubs, clientOptions(namespace));
+    // Typed service client: dispatches operations by method reference on the service interface.
+    NexusServiceClient<GreetingNexusService> greetingClient =
+        nexusClient.newNexusServiceClient(GreetingNexusService.class, ENDPOINT_NAME);
+
+    demonstrateExecute(greetingClient);
```
Quinn-With-Two-Ns: Please inline all these functions, they only have one caller and make the code harder to follow

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
```diff
@@ -0,0 +1,293 @@
...
+    // execute(...) starts the operation and blocks until it completes, returning the result in one
+    // call (equivalent to start(...).getResult()). Used here on the synchronous 'greet' operation.
+    GreetingOutput executed =
+        nexusClient.execute(
+            GreetingNexusService::greet, new GreetingInput("execute"), basicOptions());
+    logger.info("execute() returned: {}", executed.getMessage());
+
+    // executeAsync(...) is the same but returns a CompletableFuture instead of blocking.
+    CompletableFuture<GreetingOutput> future =
+        nexusClient.executeAsync(
```
Quinn-With-Two-Ns: I would show `execute` here to keep it simpler

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
```diff
@@ -0,0 +1,293 @@
...
+
+    // A single NexusClient is the entry point: it serves Visibility queries (list/count) and
+    // produces service-bound clients.
+    NexusClient nexusClient = NexusClient.newInstance(stubs, clientOptions(namespace));
+    // Typed service client: dispatches operations by method reference on the service interface.
+    NexusServiceClient<GreetingNexusService> greetingClient =
+        nexusClient.newNexusServiceClient(GreetingNexusService.class, ENDPOINT_NAME);
+
+    demonstrateExecute(greetingClient);
+    demonstrateCancel(greetingClient);
```
Quinn-With-Two-Ns: I don't think we demonstrated `cancel` or `terminate` in other SDK language samples. Not strongly opposed, but it is a lot? Whatever we decided to do we should be consistent

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
Quinn-With-Two-Ns: Another thing we discussed with Product we want to show is getting a handle to an operation that has already been started, would be good to show that when we show execute

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
```diff
@@ -0,0 +1,293 @@
...
+    demonstrateTerminate(greetingClient, client);
```
Quinn-With-Two-Ns: I'd also demonstrate start

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
```diff
@@ -0,0 +1,293 @@
...
+    demonstrateVisibility(nexusClient);
+    demonstrateClientOptionsAndInterceptors(stubs, namespace);
```
Quinn-With-Two-Ns: Interceptors are a very advanced option I would leave them off this sample

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
```diff
@@ -0,0 +1,293 @@
...
+        nexusClient.start(
+            GreetingNexusService::startGreeting,
+            new GreetingInput("to-cancel-" + RUN_ID),
+            basicOptions());
+    logger.info("Started 'to-cancel' id={}, requesting cancellation", handle.getNexusOperationId());
+    handle.cancel("standalone-nexus sample: cancel demo");
+    logger.info(
+        "Operation id={} final status: {}",
+        handle.getNexusOperationId(),
+        awaitTerminalStatus(handle, Duration.ofSeconds(10)));
```
Quinn-With-Two-Ns: We should not be encouraging users to manually poll describe, users can just get the result of the operation through the handle

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
```diff
@@ -0,0 +1,293 @@
...
+
+// Sample client for standalone Nexus operations — operations started and managed directly by a
+// client rather than from within a workflow. Each capability is shown in its own method, called in
+// turn from main(): executing an operation and reading its result, cancelling and terminating an
+// operation, querying operations via Visibility, and configuring client options and interceptors.
+public class StandaloneClientStarter {
+  private static final Logger logger = LoggerFactory.getLogger(StandaloneClientStarter.class);
+
+  // Must match the Nexus endpoint configured on the server (see README).
+  public static final String ENDPOINT_NAME = "nexusstandalone-endpoint";
```
Quinn-With-Two-Ns: nit
```suggestion
  public static final String ENDPOINT_NAME = "nexus-standalone-operation-endpoint";
```

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
```diff
@@ -0,0 +1,293 @@
...
+
+  // A per-run suffix appended to workflow-backed operation names so their backing workflow IDs are
+  // unique on each run. Without this, re-running against the same server (no restart) would reuse
+  // deterministic workflow IDs from the previous run and collide.
+  private static final String RUN_ID = UUID.randomUUID().toString().substring(0, 8);
```
Quinn-With-Two-Ns: Run ID already has a meaning in Temporal as a server generated UUID I wouldn't overload the term here. No strong opinion on the term other then not using an existing term

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
Quinn-With-Two-Ns: using the workflow operation specifically. I would show starting and then getting the result in two steps. That is by far the most common use of Nexus so we should show that IMO

core/src/main/java/io/temporal/samples/nexusstandalone/README.MD
```diff
@@ -0,0 +1,99 @@
...
+```bash
+temporal server start-dev \
+  --dynamic-config-value nexusoperation.enableStandalone=true \
+  --dynamic-config-value history.enableChasmCallbacks=true
+```
+
+Create the namespace and the Nexus endpoint:
+
+```bash
+temporal operator namespace create --namespace default
```
Quinn-With-Two-Ns: Not needed

Quinn-With-Two-Ns: with the dev server

core/src/main/java/io/temporal/samples/nexusstandalone/README.MD
Evanthx: Done.

core/src/main/java/io/temporal/samples/nexusstandalone/README.MD
Evanthx: Added 1.7.2-standalone-nexus-operations. What is the procedure for finding and upgrading places that specify versions?

core/src/main/java/io/temporal/samples/nexusstandalone/service/GreetingIds.java
Evanthx: I get why you are saying this, but I didn't want to have two unconnected places (one class in the nexusstandalone/handler package and one in the nexusstandalone package) constructing the names individually. They both already import this package, and this way the code that constructs the name isn't duplicated, and can easily be connected by a reader of the sample code through this method.

So it's not that this is complicated or can't be inlined - it could, easily - but this removes the duplicate code that could go out of sync (in a real world example, anyway) and adds connective tissue linking the two places.

core/src/main/java/io/temporal/samples/nexusstandalone/service/ClientOptions.java
Evanthx: I can if you prefer - but then it is in two places, though I know that is done as well. The real reason I split it out is that the main method of StandaloneClientStarter started to get big, and it made it harder to parse. I have to say I like this a lot better than duplicating the code inside so I'd prefer to leave it - but I also recognize that the other samples do that duplication, so this is not the hill I'll die on. :-) So I'm replying back with that preference - but if you disagree I'll accept and go inline it.

core/src/main/java/io/temporal/samples/nexusstandalone/service/ClientOptions.java
Evanthx: I went this way when using the toml file as I found this in other sample code calls when trying to figure out how to load in the toml file. So I guess both are pre-existing patterns?

We can unify to one maybe. I think the big difference as poked at these just now is that I don't have TEMPORAL_CONFIG_FILE set so didn't even realize until now the other samples relied on that environment variable. Without it set it defaults to ~/Library/Application Support/temporalio/temporal.toml which doesn't exist, and I don't see TEMPORAL_CONFIG_FILE anywhere in this codebase, not even the readmes, so how do people know to set it?

I like this pattern much better. It just works!

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
Evanthx: You are talking about demonstrateExecute, demonstrateCancel, etc?

I would disagree - I had them all inlined originally, and I find this MUCH clearer. When all inlined, it just became a huge method that looked very daunting and was really difficult to parse, this is bite sized chunks that are much easier to understand.

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
Evanthx: Execute is just above on line 66, I figured I'd show both ways!

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
Evanthx: Good point. I have it starting in the cancel and Terminate examples, but it's not called out so is therefore not at all obvious that it is present!

core/src/main/java/io/temporal/samples/nexusstandalone/README.MD
Evanthx: Removed

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
Evanthx: Changed

core/src/main/java/io/temporal/samples/nexusstandalone/service/GreetingIds.java
Evanthx: Removed

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
Evanthx: Removed

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
Evanthx: Done

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
Evanthx: Done

build.gradle
```diff
@@ -21,7 +21,7 @@ subprojects {
     ext {
         otelVersion = '1.30.1'
         otelVersionAlpha = "${otelVersion}-alpha"
-        javaSDKVersion = '1.35.0'
+        javaSDKVersion = '1.36.0-SNAPSHOT'
```
Quinn-With-Two-Ns: Can't merge this PR until `1.36.0` is released

build.gradle
```diff
@@ -10,7 +10,7 @@ subprojects {
     apply plugin: 'com.diffplug.spotless'
 
     compileJava {
-        options.compilerArgs << "-Werror"
+         options.compilerArgs << "-Werror"
```
Quinn-With-Two-Ns: uneeded whitespace change

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
```diff
@@ -0,0 +1,216 @@
...
+    demonstrateExecute(greetingClient);
+    demonstrateStartAndCancel(greetingClient);
+    demonstrateStartAndTerminate(greetingClient, client);
+    demonstrateVisibility(nexusClient);
+  }
+
+  // ─────────────────────────────────────────────────────────────────────────────────────────────
+  // execute — run a standalone Nexus operation and return its result.
+  // ─────────────────────────────────────────────────────────────────────────────────────────────
+  private static void demonstrateExecute(NexusServiceClient<GreetingNexusService> nexusClient)
```
Quinn-With-Two-Ns: Did we add a samples showing how to get a handler from an operation ID?

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
```diff
@@ -0,0 +1,216 @@
...
+  private static NexusClientOptions clientOptions(String namespace) {
+    return NexusClientOptions.newBuilder().setNamespace(namespace).build();
+  }
+
+  /** Builds the per-call options used to start a Nexus operation. */
+  private static StartNexusOperationOptions basicOptions() {
+    return StartNexusOperationOptions.newBuilder()
+        // Required: a namespace-unique operation ID. The SDK never generates one for you, so you
+        // must supply your own (a UUID here).
+        .setId(UUID.randomUUID().toString())
```
Quinn-With-Two-Ns: We should set a meaningful ID here , we should not just set the ID to random, it is bad pattern we don't want users or AI to copy.

Quinn-With-Two-Ns: I'd give each operation an ID relevant to what we are trying to demo/show

build.gradle
Evanthx: Fixed!

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
Evanthx: Done

core/src/main/java/io/temporal/samples/nexusstandalone/StandaloneClientStarter.java
Evanthx: I got the handle but not by ID! Added. Leaving this thread open - close it if you approve of what I added? It is now in demonstrateExecuteAndGettingHandleById

Quinn-With-Two-Ns (approved):
Approving, but this cannot merge until the Java SDK support is released

jsundai:
Checking in on this! Is it ready to be merged?

Quinn-With-Two-Ns:
@jsundai Yep! let me just resolve a merge issue then I will merge this

